PR #112, opened 2014-12-02
Update jquery.responsiveimages.js

s-leger:
Add options for container and skip_invisible
according : http://www.appelsiini.net/projects/lazyload
Webkit browsers will report images without width and height as not .not(":visible"). This causes images to appear only when you scroll a bit. Either fix your image tags or set skip_invisible to false.

Better handling of top/bottom left/right position of element
inspired by http://www.appelsiini.net/projects/lazyload
to allow onepage horizontal layouts with jquery.ascensor.js
http://kirkas.ch/ascensor/#Home

Allow external call to $.fn.responsiveimages.update()
eg on ascensor page transition

### DIFF
--- a/Resources/Public/JavaScript/Libs/jquery.responsiveimages.js
+++ b/Resources/Public/JavaScript/Libs/jquery.responsiveimages.js
@@ -17,7 +17,9 @@
 					720: 'medium',
 					940: 'large',
 					1140: 'bigger'
-				}
+				},
+				container:window,
+                                skip_invisible: true
 			},
 			options = $.extend({}, defaults, options),
 			$window = $(window),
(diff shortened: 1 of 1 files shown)


webian:
Hi @benjaminkott and @s-leger !
Correct me if I'm wrong but this edit of jquery.responsiveimages.js is uneffective cause EXT:bootstrap_package loads the min version: jquery.responsiveimages.min.js.

s-leger:
Hi,
According :
https://github.com/benjaminkott/bootstrap_package/issues/131
the edit is not active by default.


Le Fri, 13 Mar 2015 10:54:25 +0100, Ian Excedo <notifications@github.com>  
a �crit:

> Hi @benjaminkott and @s-leger !
> Correct me if I'm wrong but this edit of jquery.responsiveimages.js is  
> uneffective cause EXT:bootstrap_package loads the min version:  
> jquery.responsiveimages.min.js.
>
> ---
> Reply to this email directly or view it on GitHub:
> https://github.com/benjaminkott/bootstrap_package/pull/112#issuecomment-78893838


-- 
____________________________________________________

Stephen  Leger
Info-graphisme    3d services
Route du Caro
SWITZERLAND  CH-1965 Saviese

Tel   :  +41 (0)27 395 24 86
Fax   :  +41 (0)27 395 24 87
E-mail  : stephen@3dservices.ch

www.3dservices.ch

____________________________________________________
This email and any files transmitted with it are confidential and
intended solely for the use of the individual or entity to whom they
are addressed. If you have received this email in error please
notify the system manager : postmaster@3dservices.ch
____________________________________________________

webian:
Hi @s-leger 
what I mean is that it has been pulled a change to
jquery.responsiveimages.js
but, by default, bootstrap_package loads:
jquery.responsiveimages.min.js

So, the change you proposed should be made also in
jquery.responsiveimages.min.js

PR #122, opened 2014-12-13
Simplify language mapping

aimeos:
Another minor improvement to simplify the language mapping. Accoring to http://huuah.com/multi-language-setup-in-typo3/, we can also use the ISO language codes as parameters instead of the language UID number. Thus, we don't need an additional mapping in RealURL.


benjaminkott:
we already improved the configuration options, its still no the best we can reach but all language settings are not configurable via constants or adapting the language template

PR #140, opened 2015-02-26
[FEATURE] allows carousel item header to be blank in the backend

webian:
Hi @benjaminkott,
these simple edits allow the header of the carousel item to be blank so the carousel can also be used as a simple image gallery.

Please merge if possible.
Thanks!


benjaminkott:
@webian in general i like the idea - but dont you think its better to build an own type for this?

benjaminkott:
@webian this is what i would like to avoid.

not-identified-records

webian:
@benjaminkott good point! I will work on it.

PR #145, opened 2015-03-18
[BUGFIX] Typoscript languagemenu fix language stdwrap

ruudsilvrants:
The language menu shows now the correct name for the uid. (uid 1 =
Deutsch and uid 2 = Dansk)


cedricziel:
Closed through #220

PR #151, opened 2015-03-23
Responsive images

s-leger:
Extends typo3 6.2-7.0+ bootstrap_package by

Automatically pre-compute real image size for each breakpoint and generate images according.
Use modified jquery.responsiveimages.js to load the right source on each target device.
Adding optional support for retina.
Responsive images are great, but providing images with the right size for each breakpoint and layout is a real pain. The main constrains are :

Devices scree sizes, including retina.
Images sizes not always use the full width of the layout (images beside text, multiple columns...)
Columns width can vary in a single template.
The layouts are changing across breakpoints.
Nested columns layouts are possible with gridelements. (We are about a gridelements columns for bootstrap extension build to take advantage of this extension.)
Here is a working solution adressing nearly every of those cases.

### DIFF
--- a/Configuration/ContentElements/Base.ts
+++ b/Configuration/ContentElements/Base.ts
@@ -1,50 +1,233 @@
 ###########################
 #### RESPONSIVE IMAGES ####
 ###########################
-tt_content.image.20.1 {
-	layout {
-		bootstrappackage {
-			element (
-				<img src="typo3conf/ext/bootstrap_package/Resources/Public/Images/blank.gif" ###PARAMS######SOURCECOLLECTION######ALTPARAMS######SELFCLOSINGTAGSLASH###>
-				<noscript>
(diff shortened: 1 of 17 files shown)


s-leger:
Published as TER extension lef_responsive_images, including a far more advanced feature set. 
- complete support for <picture> and <srcset>
- better responsive-images including throttle and exact media query detection as jquery .width() fail to do so.
- fixes in images layout (textpic) and image-wrap sometimes preventing mouse click.

benjaminkott:
@s-leger is this still valid or can we close the pull request?

s-leger:
No more valid, you can close it  
it's an extension (lef_responsive_images) available from TER and on github
https://github.com/s-leger/lef_responsive_images

benjaminkott:
thx, closed it

PR #152, opened 2015-03-23
[Bugfix] fixes missing localization support in FalViewHelper.php

brrrrrrrt:
the render function in FalViewHelper.php is missing localization
 -> localized background images in the Carousel are not shown in the frontend,
instead the background from the default language is shown.

the changes are "inspired" by sysext/frontend/Classes/Page/PageRepository.php

i am not sure if its needed to include the _PAGES_OVERLAY_UID check, that part may be omitted if the helper will never gets used on page like records.


benjaminkott:
hey @brrrrrrrt, thx for the corrections!

brrrrrrrt:
you are welcome, i have to say thank You for this great extension!

PR #163, opened 2015-04-11
[BUGFIX] fix for HTML5 markup validation

webian:
Please merge if ok, tnx!

### DIFF
--- a/Configuration/TypoScript/constants.txt
+++ b/Configuration/TypoScript/constants.txt
@@ -16,12 +16,14 @@
 ############
 page {
 	logo {
-		# cat=bootstrap package: basic/110/file; type=string; label=Logo: Leave blank to use website title from template instead
+		# cat=bootstrap package: basic/110/100; type=string; label=Logo: Leave blank to use website title from template instead
 		file = EXT:bootstrap_package/Resources/Public/Images/BootstrapPackage.png
-		# cat=bootstrap package: basic/110/height; type=int+; label=Height: The image will not be resized!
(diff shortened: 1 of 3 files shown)


webian:
Hi @benjaminkott!
If I got what you say, you don't like the word "logo" after the sitetitle.
We know that alt attribute is used to describe the image expecially for screen reader or when the element cannot be rendered so I added the word "logo" to let user understand that it's in fact the logo of the website. Leaving only the site title did look a little misleading to me.

Btw, you're right saying to make it configurable, so I made it with the last commit.
Does it sound good? :)

webian:
I thought about that (use a constant) but I think that only an advanced user (aka an administrator) will want to hack the value of the alt attribute of the logo.
So I think that it's better to not add another constant that could add more confusion to a normal user. Also the package remain more clean.
But if you think that a constant is necessary I will add, no problem!

benjaminkott:
@webian if you want to change the logo you will anyways need to open the constant editor in most cases, if we just add it next to the logo path it should ne not a big deal.

webian:
Hey @benjaminkott, what I mean is that probably a normal user won't be interested in edit an alt attribute.
But I agree that's not a big deal and I added a constant to overwrite the default alt text.
What do you think about it?

benjaminkott:
@webian looks fine for me, will merge it move everything else into a followup
thx as always, hope youre at the devdays - ill owe u some beers already

webian:
Tnx @benjaminkott 
well... I owe you many more! :)

PR #168, opened 2015-04-19
[FEATURE] CSS source mapping

webian:
Hi @benjaminkott,
this feature enable CSS source mapping (https://developer.chrome.com/devtools/docs/css-preprocessors) very useful to debug the generated CSS from Less. It's enabled if new constant 'cssSourceMapping = 1' and requires 'config.compressCss = 0'.

Thanks for your attention :)

P.S.:
Can you also take a look at the other pull requests: https://github.com/benjaminkott/bootstrap_package/pull/163 and https://github.com/benjaminkott/bootstrap_package/pull/167?

Issue #162, opened 2015-04-10
Option to disable Less compile.


webian:
Hi @benjaminkott,
is there a way to disable the compile of Less into css files?

What I mean is a way to have:
```<link rel="stylesheet/less" type="text/css" href="styles.less" />```
instead of the compiled:
```<link rel="stylesheet" type="text/css" href="typo3temp/bootstrappackage/lessphp_b54a14e36e0e906b7bbec1e2282798c39a319f73.css" media="all">```

If not, it could be useful in development because then we can use less.js (http://lesscss.org/#download-options) and work with Less files in "real-time" instead of waiting the re-generation of the page on each edit of a Less file.

Waiting for your opinion... bye!

**Comment by benjaminkott** [comment]
Hey @webian, ive included an option in the extension setting where you now can disable the less processing. Make sure you also exclude the less file from concatenation and compressing.

Personally i would include do it a bit different and go for something like https://prepros.io/ that have already live refresh in its toolset. You would just include a normal css file and disable the concatenation and compressing for that file so the tool can find the css file and replace it as soon as you make changes to your styles.

But it does not hurt to have that option. Thx a lot for the suggestion.
Greetings Benji

**Comment by webian** [comment]
Hi @benjaminkott, I'm checking out prepros.io, looks useful, thank you for the suggestion and for this feature!

**Comment by webian** [comment]
Hi @benjaminkott,
I checked Prepros and I'm trying to figure out the best way to use it with TYPO3+bootstrap_package in development (not in production obviously).

I succeded in using it with live preview while editing the less file but I don't know how to include the variables without hacking bootstrap_package.

More in detail, I use my own service in my own extension to build a file "variables.less" that is compiled getting variables from constant. Then I should @import variables.less just after the Bootstrap "variables.less". Like:
```
// Core variables and mixins
@import "variables.less";

// here goes my own generated file with variables from constants
@import "myvariables.less";

@import "mixins.less";
...
```
Do you have any suggestion for me to solve this puzzle without hacking Bootstrap files?
If I can ask, how to you handle your less development?

PR #169, opened 2015-04-20
[FEATURE] New advanced constant to enable/disable the use of Typoscript constants as Less variables

webian:
Hey @benjaminkott,
this is another patch (the last one) needed to use an external Less processor like Prepros.
This doesn't change the default behaviour of EXT:bootstrap_package but allows advanced users to set a constant to disable the use of Typoscript constants as Less variables.
In this case the user must customize the default Bootstrap variables directly in Less.

Related to:
https://github.com/benjaminkott/bootstrap_package/issues/162
https://github.com/benjaminkott/bootstrap_package/pull/168


benjaminkott:
thx @webian already wanted to implement this one - need to make some adjustments will do it in a followup

PR #170, opened 2015-04-21
[FEATURE] New advanced constant to enable/disable CSS source mapping

webian:
Hi @benjaminkott,
here is the rewrite of this feature (ref. https://github.com/benjaminkott/bootstrap_package/pull/168).

This feature enable CSS source mapping (https://developer.chrome.com/devtools/docs/css-preprocessors) very useful to debug the generated CSS from Less. It's enabled if new constant 'plugin.bootstrap_package.settings.cssSourceMapping = 1'. It will also automatically disable CSS compression.

### DIFF
--- a/Classes/Service/CompileService.php
+++ b/Classes/Service/CompileService.php
@@ -53,6 +53,22 @@ class CompileService {
 				);
 
 				$settings = ($GLOBALS['TSFE']->tmpl->setup['plugin.']['tx_bootstrappackage.']['settings.'] ?: array());
+				if ($settings['cssSourceMapping']) {
+					// enable source mapping
+					$optionsForSourceMap = array(
+						'sourceMap'         => true,
+						'sourceMapWriteTo'  => GeneralUtility::getFileAbsFileName('typo3temp/bootstrappackage') . '/bootstrappackage.map',
+						'sourceMapURL'      => '/typo3temp/bootstrappackage/bootstrappackage.map',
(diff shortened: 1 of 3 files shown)


benjaminkott:
@webian had a look at this today, having problems with the sourcemap is not generated per less file and the less sources are in protected directorys so i am unable to inspect them. what do you think?

webian:
Hey @benjaminkott,
when this feature is enable a single file
```
/typo3temp/bootstrappackage/bootstrappackage.map
```
is created with the source mapping of every Less file compiled by Less.php
So we don't need a sourcemap for every less file (if I got your doubt).

The less sources are protected by the .htaccess in folder
```
/typo3conf/ext/bootstrap_package/Resources/Private/
```
but there are ways to disable this, by instance with a directive
```
AllowOverride None
```
in the Apache config of the development machine or by IP.

At the end we will have a Style inspector in Development Tools of Chrome like this:

Style inspector

where a CTRL+click on a CSS value will bring us to the Source panel in the exact line of the Less file where that value has been set.

Source panel

We can then edit it and save the Less file and, if we have set up a watcher and a Less compiler (like Prepros by instance), we can see our changes take effect nearly in real-time. :+1:

webian:
Hi @benjaminkott!
did you check my reply? If you still have problems with sourcemap just let me know and I'll try to help.

benjaminkott:
@webian sorry for the delay, i am merging this for now - we will see how its doing after people starts using it.

PR #178, opened 2015-05-12
[TASK] Added missing description to Bootstrap elements

randomresult:
Added description for Text and Icon, Carousel, Acordion, Panel and List Group


benjaminkott:
@randomresult du bist geil! Danke dir!

PR #190, opened 2015-06-08
Fine tune jquery.responsiveimage.js

s-leger:
Hi everybody !

@benjaminkott  Your (or whoever does it) job on this plugin is awsome guys, it's a real pleasure to make some little enhancement 

Line 119  
Add .bk2k namespace to window.on(load) event
Allow re-init plugin when page content does change eg by ajax call. The event namespace is intentionnaly left without .responsiveimage
to allow scroll and resize events to be disabled before ajax content changes without killing the window.load event

```javascript
// Re-init plugin with new content after ajax call
$(window).trigger('load.bk2k');

// Remove events listeners before ajax call
$(window).off('bk2k.responsiveimage');

// Force images update on eg tab change, carousel change... 
$(window).trigger('scroll.bk2k.responsiveimage');
```

Line 84 
There is no more callback on image loaded witch is an issue for content witch depend on layout. (eg bootstrap scrollspy) 
To adress this, one could fire a user event to trig callbacks when images load / change. 
Not optimal since it does fire when image start to load but on the other hand it is damn simple and does work on any browser.

See "Caveats of the load event when used with images", https://api.jquery.com/load-event/
Another cross brother approach to catch image load event exist but are far from being simple see
https://github.com/desandro/imagesloaded/blob/master/imagesloaded.js 

The event is meant to be catched with a setTimeout based handler since many calls are possible, this way make a kind of throttle.

```javascript
var loadedTimeout;
$(window).on('loaded.bk2k.responsiveimage', function(){
	clearTimeout(loadedTimeout);
	loadedTimeout = setTimeout(function(){
		// whatever you want to do 
		refreshScrollSpy();
		}, 200);
	});
```

Line 80 
Source and data-attrib 
When plugin init, all data-attrib are put into options, including sources ones.
So there is no need to get image data-attrib but just this.options[this.attrib]

### DIFF
--- a/Resources/Public/JavaScript/Libs/jquery.responsiveimages.js
+++ b/Resources/Public/JavaScript/Libs/jquery.responsiveimages.js
@@ -50,7 +50,7 @@
 		var old_attrib = this.attrib;
 		$.each(this.options.breakpoints, function (breakpoint, datakey) {
 			if (containerWidth >= breakpoint) {
-				attrib = "data-" + datakey;
+				attrib = datakey;
 			}
 		});
 		if (old_attrib !== attrib) {
@@ -77,11 +77,11 @@
 		if (!this.options.preload && this.options.skip_invisible && this.$element.is(":hidden")) return;
 		var inview = this.options.preload || this.inviewport();
 		if(inview){
-			var source = this.$element.attr(this.attrib);
-			source = source || this.$element.attr("data-src");
+			var source = this.options[this.attrib] || this.options["src"];
 			if (source) {
 				this.$element.attr("src", source);
 				this.$element.css("opacity", 1);
+				$(window).trigger('loaded.bk2k.responsiveimage');
 			}
 		}
 	};
@@ -116,7 +116,7 @@
 
 	// RESPONSIVE IMAGES API
 	// =====================
-	$(window).on('load', function() {
+	$(window).on('load.bk2k', function() {
 		$('img.lazyload').each(function() {
 			var $image = $(this);
 			var data = $image.data();
--- a/Resources/Public/JavaScript/Libs/jquery.responsiveimages.js
+++ b/Resources/Public/JavaScript/Libs/jquery.responsiveimages.js
@@ -14,6 +14,7 @@
 		this.$element	= $(element);
 		this.options	= $.extend({}, ResponsiveImage.DEFAULTS, options);
 		this.attrib 	= "src";
+		this.loaded		= false;
 		this.checkviewport();
 	};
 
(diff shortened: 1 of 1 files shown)


benjaminkott:
@s-leger hey, tested this and seems to work fine so ive merged the changes. thx again a lot for your effort.

I´ve updated the indentions and generated a minified version in https://github.com/benjaminkott/bootstrap_package/commit/cb37fbd797a05c3de7c2f24f1f9ba03c11ced065. Meanwhile i´ve also build grunt tasks for the setup so generation of minified versions is now standardized.

s-leger:
Ok, here is the documentation on how to use this version of the plugin. Feel free to make my english better so everyone is able to understand ;) .

It is now possible to "preload" (load image even when not visible) simply by adding a **data-preload="true"** attribute. Some adjustements in typoscript setup may be needed to do so.


#### Init plugin by hand eg after content change by ajax call.
```javascript
$('img.lazyload').responsiveimage();
```

**NOTE :** 
*The selection $('img.lazyload') is put into a local cache to speed up update events.  Any call to the  plugin refresh the cached selection. You should always use a jQuery selector containing the full set of images.*
 **Don't ever do this as it will make the plugin to stop updating other images than #image1.**
```javascript
 $('#image1').responsiveimage() 
```


#### Trigger an update when layout change eg with carousel or tab.
```javascript
$('img.lazyload').responsiveimage('unveil');

// For sample
function unveil(){
  $('img.lazyload').responsiveimage('unveil');
 }
// use unveil to update Carousel 
$('div[id^=carousel-]').on('slid.bs.carousel', unveil);
// use unveil to update Tab
$('a[data-toggle="tab"]').on('shown.bs.tab', unveil);
```


#### Execute a function when image(s) load, eg to update scrollspy.
The event is meant to be catched with a setTimeout based handler since many calls are possible, this way make a kind of throttle. You may want to adjust the timeout.

```javascript
var loadedTimeout;
$(window).on('loaded.bk2k.responsiveimage', function(){
    clearTimeout(loadedTimeout);
    loadedTimeout = setTimeout(function(){
        // whatever you want to do.
        // scroll spy refresh as sample 
        $('[data-spy="scroll"]').scrollspy('refresh');
        }, 200);
    });
```

benjaminkott:
@s-leger thx for your work. There is currently an issue we discovered on slack. When no lazyload images are on the page the var viewport is not initialized and update is not available.

I´ve did a quick workaround for now in https://github.com/benjaminkott/bootstrap_package/commit/10889ca3fb627b639b542b1d884b7bcbefc12cf6. Maybe you can have a look again?

PR #199, opened 2015-06-22
Fix 'overridden' typos

piccaso:
I know this is pedantic, but i keep seeing these...
Can't help myself, I just had to propose a fix.


benjaminkott:
@piccaso thx a lot - will try to be proofread better in the future ;)

PR #210, opened 2015-07-28
[BUGFIX] Use always $GLOBALS[TCA]

maddy2101:
this breaks the requestUpdate ctrl string in tt_content, see https://forge.typo3.org/issues/68517


neufeind:
big +1 :-)

PR #211, opened 2015-07-31
Responsive Images plugin documentation

s-leger:
Hi,
Just added in depth documentation for jquery.responsiveimages.js 
Not sure about location of this one into the doc, but put into administrator.

Note : javascript codeblock markup seem to be unsupported by github, but is the code generated by sphynx. So expect to be correct in the typo3 doc.


benjaminkott:
@s-leger thx a lot for documenting this

Issue #213, opened 2015-08-04
internal server error after installing v6.2.14 on PHP 5.4


brrrrrrrt:
it seems **class.ext_update.php** is not compatible with PHP 5.4?

after clicking to update the extension, the extension manager breaks / shows an empty page.
no errors logged, either in typo3 log, php log or apache error log - extensionmanager becomes unusable as it shows an empty page only.

after renaming class.ext_update.php the extension manager works again (but no db update has been done)

i am not sure whats the exact cause, but it must be within class.ext_update.php (i know PHP 5.4 is far from up2date, but may be still around...)

benjaminkott:
@brrrrrrrt yeah, i see the problem - we currently going to make of atleast PHP 5.5 because 14th September 2015 is EOL for 5.4. I am unsure about how to handle this.

benjaminkott:
@brrrrrrrt will check the code this evening and will provide a PHP 5.3 patch.

neufeind:
Customer stepped into the same trap. Broken extension-manager without any sign of reason *sigh* Would be great if you could get this one-liner in

cedricziel:
Resolved by #216 

Please update to git master or cherry-pick the commit.

PR #216, opened 2015-08-11
[BUGFIX] Make class.ext_update.php work on PHP <5.5 as well

neufeind:
TYPO3 6.2 still runs on PHP 5.3/5.4. Using ::class is only
possible starting from PHP 5.5.

Resolves: #213


cedricziel:
Old syntax still works but rather logs a deprecation.

neufeind:
It logs a deprecation? How? For TYPO3 it's one and the same "string" that is provided to the function.

cedricziel:
You're right.

PR #232, opened 2015-09-01
Add useful RTE buttons

phvt:
The insertcharacter and findreplace buttons added here are in the htmlArea RTE's default "Typical" configuration, so editors may expect to see them.


cedricziel:
Thank you for the contribution!

For the next one, It would be nice if you could stick to the rules defined in the `.editorconfig` (or let your editor manage it).

Fixed by a50ed6e474c41dc0109d07eb858c0be6e48c8ff5

PR #233, opened 2015-09-01
Carousel header should not be required

phvt:
The carousel header field should behave like tt_content.header, which is not a required field.

### DIFF
--- a/Configuration/TCA/tx_bootstrappackage_carousel_item.php
+++ b/Configuration/TCA/tx_bootstrappackage_carousel_item.php
@@ -310,7 +310,7 @@ return array(
             'config' => array(
                 'type' => 'input',
                 'size' => 50,
-                'eval' => 'trim,required'
+                'eval' => 'trim'
             ),
         ),
         'header_layout' => array(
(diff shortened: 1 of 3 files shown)


phvt:
This is incomplete -- I'm working on changes to Partials/ContentElements/Carousel/* so that empty header tags are not rendered.

webian:
Hi @phvt,
please also check conversation in https://github.com/benjaminkott/bootstrap_package/pull/140 about carousel header not required.

phvt:
Thanks @webian,

What if the layout for the carousel header could be set to Hidden (not rendered), as with tt_content.header_layout?

webian:
Sounds good to me but you should also ask to @benjaminkott.
( If you don't know we have a Slack channel dedicated to EXT:bootstrap-package: https://typo3.slack.com/messages/bootstrap-package/ )

benjaminkott:
@phvt if i get this correctly you want to have a variant where you only put out an image and the textfield without the header, is this correct?

benjaminkott:
with https://github.com/benjaminkott/bootstrap_package/commit/e85c213ff8ef6545130db231b7bcfa7cda2a8011 and the introduction of automatic lookup of templates https://github.com/benjaminkott/bootstrap_package/commit/42a676dbc781dc3deb47ef00fb9e1ba12d5fd537 the carousel should be flexible enough to handle custom types if nessesary.

PR #234, opened 2015-09-01
[TASK] Make nav hover text color editable

juschue:


benjaminkott:
this one will not make it into the bootstrap_package since we are trying to keep it as close to bootstrap itself as possible

PR #235, opened 2015-09-01
Backend layouts - add border column to more layouts

phvt:
The border column in bootstrap_package is highly useful for the Carousel and other full-width content, even on subpages. This pull request adds the border column to the layouts that lacked it, except for default_clean.


gilbertsoft:
@benjaminkott what do you think about this? should I do a rebase?

benjaminkott:
@gilbertsoft i´ll take a look - still undecided

benjaminkott:
@phvt @gilbertsoft i´ve recreated the patch since rebasing was to difficult, also images changed meanwhile to svgs. I´ve added it to all layouts except simple - that one will also drop footer columns later.

Issue #71, opened 2014-10-31
Language menu


SventB:
Can you please add the possibility to disable the language menu with a TypoScript constant, like you can disable e.g. the copyright text?

Can you please add a language menu with flags?

If you agree with me, that this is useful, I could implement this in my fork and you merge this later.

benjaminkott:
Thanks for your report, i´ve created a new epic to collect all language related issues, or areas that could need improvement: https://github.com/benjaminkott/bootstrap_package/issues/510

Will close this issue for now, but will be referenced in the epic.

Issue #30, opened 2014-07-30
Language menu shows non existent languages


justfortherec:
By default bootstrap_package shows a language menu with English, Danish and German. This is the case even when there is no additional languages configured in the CMS.

By default the language selector should only be visible if there is at least one second language and no non existent menus should be displayed. 

My guess is that languages used in the TYPO3 Introduction package are hardcoded in typoscript of bootstrap_package. This configuration should not be part of bootstrap_package but of Introduction package (or whatever extension uses needs this feature).

benjaminkott:
while the problem is still existend, it can now be configured with constants and and we provided more adaptable templates, still typoscript configuration needs to be made. as soon as we can see a light at the end of the tunnel we will try to provide a better solution.

PR #236, opened 2015-09-01
Language menu auto-hides if there are no sys_language records

phvt:
The menu itself does not auto-configure for actual language records in the website, but at
least now it won't show up when bootstrap_package is installed on a site with no language records.

See discussions in #30, #71 and #122


cedricziel:
We have a conflict here - can you rebase your changeset please?

benjaminkott:
@phvt auto hiding languages will provide various issues on multisites - we are currently not able to determine what languages are really available for a site (without parsing the whole tree). i will close this pr for now since we made meanwhile some useful additions for language handling, so it can easily be turned off or configured.

PR #237, opened 2015-09-03
Added classes to containers useful to better select them

webian:


phvt:
Perhaps instead, there should be IDs like colpos-0, colpos-1, ... colpos-N? Using IDs would fit with css_styled_content's default wrapping with IDs for each content element. I'm worried about "classitis" here.

#235 touches many of these files.

webian:
I thought about adding classes or IDs for colpos-N while I worked on this but I left them out of this PR because I think that it's a different thing.
This commit wants to generically mark subnav and content wrappers.
Marking colpos-N could be also useful but it's more specific.
Also keep in mind that subnav doesn't have a colpos-N.

PR #254, opened 2015-10-28
[TASK] If addititonal parameters are provided they should be added to…

DennisRheinfelder:
… the embeded url.


cedricziel:
You pull-request from an older state. Please rebase, force push and post a comment so we know. Cheers!

PR #262, opened 2015-11-21
Force preload images to allow print

s-leger:
Force image preload even when not visible on page to allow print. 
Use lazyload js plugin, but disable inviewport check so all images are loaded right after page load.

Adds config.preloadImages constant to enable this feature on site/per page basis.


benjaminkott:
@s-leger thx a lot

PR #264, opened 2015-12-03
[BUGFIX] Correct path for icon-font-path variable

Outdoorsman:


benjaminkott:
include paths have been double checked and are working fine, i think this one is obsolete

PR #273, opened 2015-12-22
master

Outdoorsman:


Outdoorsman:
This is especially necessary when you are doing SEO work. The title for Google should definitely be longer than something like "Home"... so we make the title "PHP Web Developer Jobs in Paris, France: Some Company" and the navigation title has to fit in a smaller space so we use "Home". This is default behavior for most menus to use the nav_title as the preferred.

benjaminkott:
@Outdoorsman thx for pointing this out! ive already added a solution for this - and forgot to adjust the templates, really sorry for that. you can set a titleField in the menu processor options that will actually do this and you do not have to care about it in the templates.

so you can instead accessing the raw data for example {item.data.title} just call {item.title}

```
  10 = BK2K\BootstrapPackage\DataProcessing\FlexFormProcessor
  10 {
    special = list
    special.value.field = pages
    levels = 7
    as = menu
    expandAll = 1
    titleField = nav_title // title
   }
```

Outdoorsman:
That works great... thanks for an even better fix :)

PR #277, opened 2015-12-30
RFC: Feature/add pdf thumbnails to imagerendering

pekue:
Using PDFs in Image / Text with Image does not render a thumbnail in the frontend. This is because file.type is 5 (document) in sys_file. I know that this proposal is not a good solution, because it doesn´t take $TYPO3_CONF_VARS['GFX']['imagefile_ext‘] into account, which it should - but it works.

### DIFF
--- a/Classes/DataProcessing/MenuProcessor.php
+++ b/Classes/DataProcessing/MenuProcessor.php
@@ -39,6 +39,8 @@ use TYPO3\CMS\Frontend\ContentObject\DataProcessorInterface;
  * Options:
  * as - The variable to be used within the result
  * levels - Number of levels of the menu
+ * expandAll = If false, submenus will only render if the parent page is active
+ * titleField = Field that should be used for the title
  *
  * See HMENU docs for more options.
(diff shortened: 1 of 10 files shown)


cedricziel:
Can you please rebase to the current master? Cheers.

s-leger:
pdf are files of type 5, so current partials/media/type should also be changed according in order to be able to use them in frontend.
```
  <f:case value="5">
    <f:switch expression="{file_0.extension}">
      <f:case value="pdf">
        <f:render partial="Media/Image" arguments="{file: file, data: data, settings: settings}" />
      </f:case>
      <f:case default="TRUE">
        <div class="danger">File rendering for type:{file_0.type} {file_0.extension} is not supported.</div>
      </f:case>
    </f:switch>
  </f:case>
```

benjaminkott:
while the change indeed sounds easy, it misses to check if ghostscript is properly installed. if ghostscript is not installed correctly it can lead to unwanted output in the frontend and it is hard to validate upfront.

benjaminkott:
asked @lolli42 about this if we can make it somehow available during installation to check it before we render the page about this, until now i will close this pr - will provide a solution when we are able to ensure processing options are available.

Issue #281, opened 2016-01-08
RTE: Classes for links


SventB:
If I want a link in button style, e.g. style "Success", I have to select in selectbox "textstyle" first "Buttons: Base" and then "Buttons: Success", because only then the a-tag gets the correct classes ("btn btn-success").

This is a little bit complicated, isn't there a better solution for that? I've tried to accomplish this with TypoScript (lib.parseFunc_RTE.externalBlocks) but didn't succeed.

SventB:
A working solution seems to be to extend the LESS styles a little bit:

```
.btn-success,
.btn-danger,
.btn-all-other-buttons {
	&:extend(.btn);
}
```

SventB:
Or better this way, for correct hover effects:
```
.btn-default {
	.btn();
	.btn-default();
}
.btn-primary {
	.btn();
	.btn-primary();
}
.btn-success {
	.btn();
	.btn-success();
}
.btn-warning {
	.btn();
	.btn-warning();
}
.btn-danger {
	.btn();
	.btn-danger();
}
```

cedricziel:
Yep. The editor won't let you attach multiple classes at once. Same goes for detaching fyi.

Do you care to make a pull-request?

PR #293, opened 2016-01-23
RTE: Classes for links

SventB:
See https://github.com/benjaminkott/bootstrap_package/issues/281


cedricziel:
Please use `git rebase -i HEAD~3` on the `master` branch.

Remove the two last lines mentioning `test1` and `revert` commit and save & close. Then force push to update the pull request.

Reason: The two commits added are useless.

cedricziel:
Hey thanks for the adjustments :)

Please note that you could've updated the original pull request by rebasing like i mentioned. Pushes to the branch then require a force push.

You may want to read that part up in the git book, as it simplifies collaboration on github a lot as pull requests don't share state PLUS it will level you up with git *a lot*.

SventB:
Yes, I have a lack in git knowledge ;) My TourtoiseGit didn't allow to rebase something.

webian:
You could give a check to Smartgit @SventB 
http://www.syntevo.com/smartgit/

cedricziel:
Use the shell, Sven. ;)

Jokes aside, TortoiseGit requires MSysGit, right? There should be a shell you can use. Along the lines there are also more adanced clients such as SourceTree (free) and SmartGit (free for OSS). I use SmartGit for repetitive tasks such as updating forks but I need the shell anyways for non-gui evironments.

SventB:
Thanks for the useful hints.
I plan to contribute to this project more, I have a list about 20 improvements I'ld like to discuss and implement. I think I will have time for that in april and then I will have a closer look to GIT.

cedricziel:
I see no use in creating knowledge silos; so please feel free to contact me about git if you need some advice. I'm not on the Slacks, but you can use the email address which is visible on my github user profile.

PR #298, opened 2016-01-28
[FEATURE] Allow to disable footer-section with Typoscript constant.

webian:


cedricziel:
So maybe we should allow displaying nothing at all. scnr

PR #302, opened 2016-01-29
[FEATURE] allow BE user to select the crop ratio of images

webian:
This feature is useful to display all the images of a gallery with the same crop but leaves the images with the original aspect when click enlarged. This crop method is different from cropping each image in the TYPO3 BE 'Image manipulation' cause in the latter also the click enlarged images will be cropped.

Initially only a square 1:1 crop is implemented but can be expanded in the future.


benjaminkott:
I like the idea, but we need to prepare these values at a different place and only pass the values to the partial for the images. Maybe its useful to have some kind of content dataprocessing or content controllers

PR #307, opened 2016-01-31
[FEATURE] Remaining PageTS templates are configurable

cedricziel:
Modifies inclusion of

* tt_content previews
* new content element wizard items

in a form that either allows them to be included
via ext_typoscript_setup or not.

This is especially helpful with overrides and migration
paths of existing sites to the bootstrap_package.


cedricziel:
I'm using this is production right now and it's fine. I'll merge it if nobody has objections :)

PR #330, opened 2016-04-14
mainmenu: add css style 'active' for shortcut page pointing to parent

synac:


synac:
Travis CI reports an issue with php 5.5. I checked it on a test server with php 5.5 and can't find any issues.

benjaminkott:
@synac hey thx for your patch will look into this later, i think we should handle this in the menu processor to avoid complex conditions, travis should be fixed now btw

gilbertsoft:
This will be solved directly in the core see https://forge.typo3.org/issues/85138

benjaminkott:
@synac @gilbertsoft added new labels, and will close the pr for now - since this should be part of the core.

PR #346, opened 2016-05-28
[BUGFIX] Prevent double escaping of menu titles

IchHabRecht:
Menu titles already get escaped by TypoScript in
\BK2K\BootstrapPackage\DataProcessing\MenuProcessor. This patch ensures
the tiles are used without escaping.


benjaminkott:
@IchHabRecht thanks!!

PR #358, opened 2016-07-06
[BUGFIX] Use field 'uploads_type' in Uploads template

marble:
The layout style of the items in the uploads list is in field '{data.uploads_type}' and not in '{data.layout}'.

### DIFF
--- a/Resources/Private/Templates/ContentElements/Uploads.html
+++ b/Resources/Private/Templates/ContentElements/Uploads.html
@@ -10,7 +10,7 @@
         <ul class="media-list">
             <f:for each="{files}" as="file" iteration="fileIterator">
                 <li class="media">
-                    <f:if condition="{data.layout} == 2">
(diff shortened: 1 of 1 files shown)


benjaminkott:
@marble i´ve double checked it, but uploads_type is not existing here, do you have fluid styled content installed?

marble:
**Uah, oh damn, you're right: ext:fluid_styled_content was activated.** Although its TypoScript was not loaded. Ah I had lost out of sight that already just LOADING fluid_styled_content interferes. I'll add a big WARNING to the docs, I think.

But now that its disabled: Why is that {data.layout} in the condition in the Uploads.html template? It seems I can't set an option somewhere. Searching ...

and watch out: Great, you're right, there it is:
image

Conclusion: You're right, this is not a bug, close the Pull Request. ATTENTION: Make sure, ext:fluid_styled_content is not activated when using ext:bootstrap_package.

Issue #366, opened 2016-08-08
[FEATURE] Add image-align-left and image-align-right


marble:
## For `text with images` and `text with media` elements we have these four options for image alignment:

![014](https://cloud.githubusercontent.com/assets/307057/17485671/9685a498-5d8e-11e6-80d0-ac46c48e0523.png)

## This is fine as long as the images are large enough to fill the complete column:

![015](https://cloud.githubusercontent.com/assets/307057/17485998/7a19b064-5d8f-11e6-9974-bd50faa6dfd0.png)

## However, if the images are very small the only option you have at the moment is to display them horizontally centered:

![016](https://cloud.githubusercontent.com/assets/307057/17486146/03c1c8d8-5d90-11e6-9806-7f97c48dd720.png)

This is a problem in real life for existing websites that have some small images and want to have them left aligned with the text for example. To address that problem I've created a patch that I'll send as a pull request.

## With that feature you have some more options in the alignment menu:
Namely there are additionally:

- above-left
- above-right
- below-left
- below-right

![017](https://cloud.githubusercontent.com/assets/307057/17486323/af1c2d68-5d90-11e6-8e38-10394f03171c.png)

## For above-left (or below-left) you get:

![018](https://cloud.githubusercontent.com/assets/307057/17486405/f95295fc-5d90-11e6-9d11-e413248c1ce1.png)

## And for above-right (or below-right):

![019](https://cloud.githubusercontent.com/assets/307057/17486461/2fcd8894-5d91-11e6-81f3-440ed68e7577.png)

## And this is the real live usage: Display a little image left justified

![020](https://cloud.githubusercontent.com/assets/307057/17486788/6d643d64-5d92-11e6-8666-edd2d2152f36.png)

## I will send a separate pull request.

The positioning is done with an extra class `image-align-left`or `image-align-right`in the `figure`-tag. In doing so the caption can be affected as well if somebody needs that. In less and css we're issuing a `left-margin: 0` or `right-margin:0` instead of the default `auto`.

## PS:

While investigating I came to the conclusion that it indeed doesn't make sense to have an alignment option `in text left` or `in text right` as that would require much more of specification to make sense in the bootstrap context.

PR #367, opened 2016-08-08
[FEATURE] Add image-align-left and image-align-right

marble:
As described at https://github.com/benjaminkott/bootstrap_package/issues/366

### DIFF
--- a/Configuration/PageTS/TCEFORM.txt
+++ b/Configuration/PageTS/TCEFORM.txt
@@ -81,7 +81,7 @@ TCEFORM {
             removeItems = 5,7,8
         }
         imageorient {
-            removeItems = 1,2,9,10,17,18
+            removeItems = 17,18
             types {
                 image {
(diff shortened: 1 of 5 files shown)


benjaminkott:
@marble 

I would like to avoid conditions for css classes in the templates and would prefer a standalone template for each variation, this makes it more easy to extend and more easy to understand, even if its duplicate code.

Also you need to generate the css files again if you made adjustments to the less files, there is still a static version that we ship that also needs to contain these stuff.

Can you please make these adjustments then i will merge it directly.

marble:
Sure, alright. Good to know what you prefer. There are so many ways to do things. I need to work on some other things first but didn't want to loose sight of this PR. I'll send and updated version and ask you to review again. Have a nice day! (No need to hurry with this one)

benjaminkott:
This one os way to long open, i finally decided not to add this option.

PR #369, opened 2016-08-12
[BUGFIX] Fix inline documentation of MenuProcessor

fsaris:


benjaminkott:
uupps @fsaris thx for fixing!

Issue #370, opened 2016-08-17
Missing addQueryString on f:link.page


webian:
In a page generated with query string parameters, like a news post or a product detail, the links generated with `f:link.page` (by instance "skip to main content" or "scroll to top") point wrongly to the list page.
This is because the `f:link.page` is missing the parameters `addQueryString="1"`.

webfacer:
i used therfore the `f:link.action`

webian:
ic but, correct me if I'm wrong, `f:link.page` with `addQueryString="1"` can generate the right link also for non extbase plugins and this way there's no need to overwrite template/partials.

webian:
Another solution could be to write directly the A tag instead of use f:link.page since the href will be statically "#content" or "#top".

benjaminkott:
Would prefer the static link, can someone create a pull request for that? Would be awesome!

benjaminkott:
@webian this seems to be resolved by your patch, thx a lot for providing it.

PR #372, opened 2016-09-14
[TASK] Replace f:link.page with static link (#370)

webian:


cedricziel:
Looks good, thank you!

PR #377, opened 2016-09-26
Responsive images generation improvement

s-leger:
Hi,
Taking some care it is possible to precompute real image size for every layout and image columns.
This is what this patch is about, also fixing low resolution screen images sizes issues.

Still dosen't take account of image borders and other frames (like well or jumbotron), but a good basis to go.

On page tempates, use "xs, sm, md, lg" vars in the lib.dynamicContent arguments with number of columns, as set on columns classes. eg: class="col-xs-6"  => xs: '6'
The biggest inherits value from smallest automatically as bootstrap.css do.

On viewHelpers managing images, do the same with the viewHelper {bk2k:imageWidth('md':8)} . This will fill a "width" variable with an array of width for each breakpoint+1.
Use this array members to set maxWidth of images generated for each breakpoint.
eg : data-small: maxWidth: width.sm


s-leger:
Wait for Art-direction branch

PR #378, opened 2016-10-03
Reduce http requests

webian:
RsaEncryptionEncoder adds its JS files with addJsFile() and bootstrap_package adds its JS files with includeJSLibs. This causes to have 2 different merged JS files in the HTML code so 2 different request.
To have a single request we change the inclusion method to includeJS since includeJSLibs is not needed in this case.

### DIFF
--- a/Configuration/TypoScript/setup.txt
+++ b/Configuration/TypoScript/setup.txt
@@ -426,10 +426,10 @@ page {
             disableCompression = 1
             excludeFromConcatenation = 1
         }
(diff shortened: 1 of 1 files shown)


webian:
I added a commit also for the CSS.

benjaminkott:
the default concatenation is turned off by default, because it is really slow and can lead to different problems. i expect that people will have their own build workflows if they really need to have a reduced number of files. 

aside from that, we need to differenciate between libaries and custom glue clode / css for that the differentiation is fine for me.

Issue #380, opened 2016-10-13
Link to shortcut page not resolved in menu


s-leger:
Hi, 
Set a page as shortcut.
In menu, the page id is used instead of shortcut target page id.
Hmenu take care of resolving links to shortcuts (and mountpoints),
but dumb link.page viewHelper dosen't, nor link.typolink.

benjaminkott:
mm.. maybe we should resolve them directly in the menu processor? what do you think?

s-leger:
Yes, 
Hmenu would also take care of mount points and all visibility stuff.
But as a drawnback, we only may get a complete url instead of uid, and so we cant no more use the link.page (dumb) viewhelper.
Maybe a separate param with a typolink in the postCObject could do the trick.

This also point out a weakness of link.typolink witch dosent take account of sutch things.

benjaminkott:
will have a look into this - lets see what we can do

benjaminkott:
ok, did a little research

1. we can easily add this directly to the processor and put a net "typolink" key or something here https://github.com/benjaminkott/bootstrap_package/blob/master/Classes/DataProcessing/MenuProcessor.php#L365-L367

2. fluid styled content does also not take this into account
3. ive found several ways how the core handles the shortcut mp checks - need a bit more digging what exactly needs to be done to get this smarter

s-leger:
> fluid styled content does also not take this into account
It's not a reason to follow bad practices :) 
Maybe they prefer server side redirection, but this wont work with static cache, ajax requests and so on.

[maybe like this ?](https://github.com/s-leger/bootstrap_package/blob/patch-10/Classes/DataProcessing/MenuProcessor.php#L160-L171)

So no need to handle this ourself, typolink will do.

benjaminkott:
currently thinking about grabbing the generated link directly from hmenu, so we do not have to care about what when and who does the job, and just make it accessible and keep the data of the record as is.

1. mountpoint creation must have menu information and cannot be handled by f:link alone because the context is missing
2. shortcut handling can be ignored because its already handled
3. we deliver the exact same data as hmenu, but in a more accessible way

webian:
I use Typoscript for the menu and I don't have problems.
What are the advantage of making a menu in Fluid?

benjaminkott:
for me some stuff is more easy to solve in fluid than typoscript, for example for mega menus or stuff like where you need a lot of markup and conditions. 

[...]

PR #381, opened 2016-10-13
Resolve shortcut links

s-leger:
Hi,
Prevent server side redirection for shortcut pages by resolving links,
to fit the standard behaviour of hmenu.
Dosen't still resolve mount points.
fix issue #380


s-leger:
Looks pretty good !
Thank you !

PR #383, opened 2016-10-13
Link handling

s-leger:
What about this ?
Dont know how to handle sections.
There is a typolink specific parameter for this, but how to handle this in the menu processor.

[typolink ref](https://docs.typo3.org/typo3cms/TyposcriptReference/Functions/Typolink/Index.html)


benjaminkott:
@s-leger typolink does not resolve the links correctly and does not care about mountpoints, please have a look at https://github.com/benjaminkott/bootstrap_package/commit/0f206e87fc9bd8fb4a70e3af85f6018a33e78332 thx again for your work

PR #387, opened 2016-10-17
RTE 2-4 text columns

### LINKED ISSUE #386: [FEATURE] Add column support for text in RTE component
Why not ?
Instead of specific content element, why don't make this a style option available for every text block ?
Maybe a set of 2, 3 and 4 columns.

s-leger:
add 2-4 columns of text in RTE

fix #386


benjaminkott:
resolved with https://github.com/benjaminkott/bootstrap_package/commit/57b7af705ed753de5ea7dcfa7e0962dd128ccd6c

PR #388, opened 2016-10-18
Single Image display doesn´t work

PaulMcGregory:
Image is not shown


benjaminkott:
@PaulMcGregory thx for your patch, i am waiting for feedback from the fluid team - this seems to be a bug in fluid standalone and should be fixed there.

PR #390, opened 2016-10-23
Art direction

s-leger:
[Available as extension ](https://github.com/s-leger/lef_responsive_images)

Currently, typo3 core only allow to set width of images without any layout relationship.

This ends up into providing images far larger than client would have needed (think about a picture in a 6 column layout rendered at 1200 / 600 / 300 px while only making 100px). Not only this consume bandwidth, but also energy to downsize images in the browsers.

Providing some care, it's possible to pre-compute real image size in nearly every layout for every breakpoint. Adding full support for correct data-attributes lazyload, css background and Art-direction.
# The cons :
- Precomputing 5 images (breakpoints +1) can be time consuming on first load of the page.
- Configuration to take account of grid system, breakpoints, and content frames.
- Adapting templates to make system aware of layout.
# The pros :
- Less bandwith
- Less energy
- Less time to load
- Automagic once configuration done, but with standards grids setups it should be ok right out of the box.
# Features
- Responsive css backgrounds
- Art direction  image and css
- Enhanced responsive carousel allowing a better height setup, and multiple images.
- Multiple images layouts : regular grid,  by row (justify, left, right) and by columns (justify, top bottom)
- Support for user defined image sizes
- Precompute real size and generate images according to the space available on any layout.
- Allow from 1 up to 10 images per row
- Support for pdf in frontend
- Html content allow to write inline css in header and js for footer (via layouts)
- Textpic image in text left and right


benjaminkott:
as mentioned before, art direction is now in place in the TYPO3 core and the bootstrap_package does fully support this, i will close this pull request since it uses a different approach. thx for your work and sorry it did not make it in.

PR #393, opened 2016-10-31
From themes bootstrap

s-leger:
Bootstrap regular and horizontal themes


benjaminkott:
the current implementation of forms are already comming with bootstrap markup out of the box, will close this one.

PR #394, opened 2016-11-02
Move alignment palette

s-leger:
[FEATURE] Move textpic alignment palette

Move alignment palette for cType textpic to appearance tab so it is the same as with images and media.


webian:
Btw @s-leger you should use https://wiki.typo3.org/CommitMessage_Format_(Git) for commits on this repo.

benjaminkott:
@s-leger @webian could you please check your setups - here its already aligned, for me you are moving it now to a wrong position

https://github.com/benjaminkott/bootstrap_package/blob/master/Configuration/TCA/Overrides/tt_content_element_textpic.php#L30
https://github.com/benjaminkott/bootstrap_package/blob/master/Configuration/TCA/Overrides/tt_content_element_textmedia.php#L30

benjaminkott:
@s-leger @webian found it, sry for confusion but i would suggest to move the media and images fields to the same position as textpic and textmedia

webian:
@benjaminkott I confirm my +1 on this PR.

webian:
@benjaminkott your option can be good to.
What matters is to uniform :)

benjaminkott:
https://github.com/benjaminkott/bootstrap_package/commit/af89966837fc70cf93083e55663d40342534b675 @s-leger @webian ok did a patch for that, i think its better to have the columns option more close visible to the images and assets itself for the editor. will close the pr in favor of that.

s-leger:
@benjaminkott Ok, sry to be late, is it possible then to have them on top of this tab ?
The point is when you have say 15 images this part mostly never visible until you scroll down.
My first idea was this palette define appearance, and we do have a tab for it, but i can live without.

benjaminkott:
@s-leger Yes, we can move it to the top, would u like to create a pr for that?

s-leger:
ok maybe late this night :)

PR #414, opened 2017-01-10
"usind" is wrong

franzholz:
if you are not usind the theme

==> using


cedricziel:
Great, thank you for your contribution, @franzholz !

PR #458, opened 2017-07-17
[BUGFIX] Fix sys_language_uid when adding item to translated tt_content

tmotyl:
### Description
In the patch https://review.typo3.org/#/c/52061/1 TYPO3 v8 now setts correct sys_language_uid on child items, when adding them directly to translated parent (tt_content).
However this patch only works when sys_language_uid is in the showitem array.

This patch adds a hidden palette which language fields. The same solution is used in the core for FAL fields.

### Steps to Validate

1. Add tt_content with acordeon item and translate it to language 1
2. see in db that translated items will have correct sys_language_uid set (1)
3. add new item to the translated tt_content
4. check in db that sys_language_uid of the child is 0
5. apply the patch, clear cache
6. redo step 3 and 4
7. see that sys_language_uid is set correctlt to 1


benjaminkott:
@tmotyl seems to be nessesary for all branches i guess. can you heep me updated when this is merged? or is it save to be merged before?

tmotyl:
Fal uses this approach, so it's safe to merge in my opinion even before the patch is merged in all branches

tmotyl:
FYI, the patch https://review.typo3.org/#/c/52061/ was just backported to v7 branch

benjaminkott:
@tmotyl thx!

PR #467, opened 2017-08-20
[FIX] Bootstrap column sizes changed

l174rd:
Bootstrap column sizes changed due to view conflicts for two column layout (should start at **col-sm**, not **col-md** in my opinion).

By the way, there's no need for **col-xs-12**, because this is the default behavior of bootstrap to give elements 100% width.


benjaminkott:
I fill close this pull request in favor of https://github.com/benjaminkott/bootstrap_package/commit/e5d9cade9925cec367bad184e90b5489280c32e5 that allows us to have the columns configurable.

PR #474, opened 2017-09-12
properly separate navbar default and inverse styles

DanielRuf:
### Prerequisites

* [x] Changes have been tested on TYPO3 8.7 LTS
* [x] Changes have been checked for CGL compliance `php-cs-fixer fix --config-file .php_cs`

### Description

This fixes the styles for the navbar (default and inverse) so they work like they should.

### Steps to Validate

1. install the introduction distribution
2. go to the homepage (pid=1)
3. see that the links have the wrong color (white)
4. change the class names to use the inverse styles and see that the colors and logo are wrong

### DIFF
--- a/Resources/Public/Less/Theme/navbar.less
+++ b/Resources/Public/Less/Theme/navbar.less
@@ -51,23 +51,27 @@
     .navbar-default-transition,
     .navbar-inverse-transition {
         .transition(all 0.27s cubic-bezier(0, 0, 0.58, 1) 0s);
+    }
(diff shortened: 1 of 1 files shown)


benjaminkott:
this one covers a specific usecase as can be seen on http://www.bootstrap-package.com/.
the problem described was introduced by an incomplete import of the testdata, no changes nessesary there.

DanielRuf:
Ah ok, I understand. Thanks for the update.

PR #478, opened 2017-10-17
BP_6-2 fix og:image setup height

revoltek-daniel:
Fix typoscript og:image height to avoid error


benjaminkott:
@revoltek-daniel thx for the fix

PR #480, opened 2017-10-20
[TASK] Avoid ambiguous "uid" error

mbrodala:
If the table "sys_category_record_mm" is promoted to a 1st level table and gets a "uid" column, the menu "Content elements for selected categories" will throw an SQL error like this:

> Column 'uid' in on clause is ambiguous

Avoid this by explicitly prefixing the table name.

Fixes # .

### Prerequisites

* [ ] Changes have been tested on TYPO3 8.7 LTS
* [ ] Changes have been checked for CGL compliance `php-cs-fixer fix --config-file .php_cs`

### Description

[Description of changes proposed in this pull request]

### Steps to Validate

1. [First Step]
2. [Second Step]
3. [and so on...]


benjaminkott:
@mbrodala thx a lot for the fix

PR #501, opened 2018-02-15
fixed path to assets

No5251:
Minor bugfix. The "Public" folder was missing in @pswp__assets-path.


benjaminkott:
@No5251 the parser is sadly not that intelligent, it depends where the source files are located and sometimes need some workarounds to get it done correctly. For now i cannot see the need to change something here since everything works fine with the default configs.

Thx a lot for pointing this out, maybe it needs more explaination somewhere.

PR #511, opened 2018-03-29
[FEATURE] Add better language support

### LINKED ISSUE #510: [EPIC] Add better language support
### Prerequisites

* [x] Can you reproduce the problem on TYPO3 8.7 LTS
* [x] Did you [perform a cursory search](https://github.com/issues?q=is%3Aissue+user%3Abenjaminkott+repo%3Abenjaminkott%2Fbootstrap_package) to see if your bug or enhancement is already reported?

### Description

The current support of languages lacks flexibility, and also comes with a defined preset of configured langauges for the current introduction package. This epic should help to collect issues so we can overview all areas that needs improvement.

### Steps to Reproduce

**Expected behavior:** [What you expected to happen]

1. I can add any language in the backend
2. The languages i´ve added in the backend are automaticly available in the frontend
3. The TypoScript configuration for added languages is automaticly generated
4. The names for the languages are taken from the language records from the backend

**Actual behavior:** [What actually happened]

1. Language names are hardcoded in the `navigation\language` template
2. TypoScript configuration has presets for en/dk/de
3. Language configuration must be added manually

### Versions

All Versions.

gilbertsoft:
Fixes #510 

### Prerequisites

* [x] Changes have been tested on TYPO3 8.7 LTS
* [x] Changes have been tested on TYPO3 9.2-dev

### Description

See issue #510 

### Steps to Validate

1. Assign a language list or `auto` to the constant `page.theme.language.languageValue` with TSCE
2. Complete `config.language.default.*` with TSCE
3. Add a new system languages and complete the existings, disabled will not be shown in auto mode
4. Check language menu at the FE


Configuration/TypoScript/setup.typoscript
```diff
@@ -223,14 +223,11 @@ page {
                     }
                 }
             }
-            40 = TYPO3\CMS\Frontend\DataProcessing\MenuProcessor
+            40 = TYPO3\CMS\Frontend\DataProcessing\LanguageMenuProcessor
             40 {
-                special = language
+                #special = language
```
cedricziel: No useless comments please

Classes/DataProcessing/LanguageMenuProcessor.php
```diff
@@ -0,0 +1,247 @@
...
+     * @param array $conf
+     * @return string JSON encoded data
+     */
+    public function getHreflangAsJson($content, $conf)
+    {
+        return $this->jsonEncode($this->cObj->data);
+    }
+
+    /**
+     * Gets the data of the current record in JSON format
```
cedricziel: Can you please drop these descriptions? They say nothing more than the method names - in fact they say less. They have to go :)

Classes/DataProcessing/LanguageMenuProcessor.php
```diff
@@ -0,0 +1,247 @@
...
+    public function __construct()
+    {
+        parent::__construct();
+
+        $this->menuDefaults['as'] = 'languagemenu';
+        $this->menuDefaults['titleField'] = '';
+    }
+
+    /**
+     * Validate configuration
```
cedricziel: Useless comment

benjaminkott:
Ok, i´ve talked to @wmdbMattes about this on how we could tackle this the best way.
we currently agreed that we will go with an class alias at this point.

The HMENU will at some point be able to render all informations we are now defining.

That would mean that we are directly starting with the bootstrap package here and are using the namespace `TYPO3\CMS\Frontend\DataProcessing\LanguageMenuProcessor` directly from the beginning in our configuration. We can do a simple check here if that class exists if not we will set an alias for the Processor delivered with the bootstrap_package with the namespace `BK2K\BootstrapPackage\DataProcessing\LanguageMenuProcessor` this will ease migrations a lot since configuration is not needed to be adaped after the official processor is available.

We should adapt the settings available in the https://github.com/bmack/typo3-sites module and define this as the return values we have. that also means we do not assign the record data of the page here for example.

Configuration/TypoScript/setup.typoscript
gilbertsoft: it's work in progress, so cleanup follows at the end

Classes/DataProcessing/LanguageMenuProcessor.php
gilbertsoft: same as before ;)

gilbertsoft:
I think it's not necessary for the core to implement a new class, the additional functions of the new processor can be integrated into the MenuProcessor core class with new properties to enabled or disable its new functionallity IMO. But then we can't do a class checking, we have to check for a T3 version.

For the moment I don't want to make new PRs to typo3-sites as long it's not sure how the config is saved, DB or yaml. I think making a solution for v8 with the view on sites for v9 is good for the moment. After the decisions are made for sites I'll continue the changes for our needs.

gilbertsoft:
@cedricziel I'll ping you if it's ready for a review

Classes/DataProcessing/LanguageMenuProcessor.php
```diff
@@ -0,0 +1,333 @@
...
+ *   includeSpacer = 1
+ *   dataProcessing {
+ *     10 = TYPO3\CMS\Frontend\DataProcessing\FilesProcessor
+ *     10 {
+ *        references.fieldName = media
+ *     }
+ *   }
+ * }
+ */
+class LanguageMenuProcessor extends MenuProcessor
```
benjaminkott: We should not depend on the currently existing menu processor, i will cause problems to keep them compatible and we do not need a lot of the stuff the menu processor does. I would suppose in that case to duplicate the code and remove everything we do not need.

Classes/DataProcessing/LanguageMenuProcessor.php
```diff
@@ -0,0 +1,333 @@
...
+            unset($this->menuLevelConfig['stdWrap.']['cObject.']['70.']['value']);
+            unset($this->menuLevelConfig['stdWrap.']['cObject.']['70.']['listNum.']);
+
+            $this->menuLevelConfig['stdWrap.']['cObject.'] = [
+                '1' => 'LOAD_REGISTER',
+                '1.' => [
+                    'languageUid.' => [
+                        'cObject' => 'TEXT',
+                        'cObject.' => [
+                            'value' => $this->menuConfig['special.']['value'],
```
benjaminkott: We should introduce a new property on root configuration level for this, in the future we will not likely have a special.value configuration here - maybe a `languages` or `languageIdentifiers` should suit better - we can map this codewise.

Classes/DataProcessing/LanguageMenuProcessor.php
```diff
@@ -0,0 +1,333 @@
...
+ * LICENSE file that was distributed with this source code.
+ */
+
+namespace BK2K\BootstrapPackage\DataProcessing;
+
+use BK2K\BootstrapPackage\Utility\LanguageUtility;
+use TYPO3\CMS\Frontend\DataProcessing\MenuProcessor;
+
+/**
+ * This menu processor utilizes HMENU to generate a json encoded menu
```
benjaminkott: Inline documentation needs cleanup later on.

Classes/DataProcessing/LanguageMenuProcessor.php
```diff
@@ -0,0 +1,333 @@
...
+    }
+
+    /**
+     * Add more language properties
+     */
+    public function prepareLevelLanguageConfiguration()
+    {
+        parent::prepareLevelLanguageConfiguration();
+
+        if ($this->menuConfig['special'] === 'language') {
```
benjaminkott: Menu is always a language menu, no need for checking this after making the processor standalone

Classes/DataProcessing/LanguageMenuProcessor.php
benjaminkott: Currently not sure about `localizedTitle` because as user i would expect to find my language wether in english or in my local language.

That would mean if i am browsing a russian website in russian - i would like to have `german` or `deutsch` in the language selector because i propably have no idea that i would be `немецкий`

benjaminkott: Output contains still the property `spacer` that is not usefull when handling language menus. Should be reomoved.

Classes/DataProcessing/LanguageMenuProcessor.php
```diff
@@ -0,0 +1,333 @@
...
+     * @throws \InvalidArgumentException
+     */
+    public function validateConfiguration()
+    {
+        $this->processorConfiguration = array_replace_recursive(
+            [
+                'special' => 'language',
+                'special.' => [
+                    'value.' => [
+                        'ifBlank.' => [
```
benjaminkott: we do not need to have all configuration need to be done through typoscript, we should evaluate if the language values are set on php side, and collect sys_languages also here. we can then just insert the generated values into `special.value`.

Classes/DataProcessing/LanguageMenuProcessor.php
```diff
@@ -0,0 +1,333 @@
...
+        $this->menuDefaults['as'] = 'languagemenu';
+        $this->menuDefaults['titleField'] = '';
+    }
+
+    /**
+     * Add loading from sys_languages in case special.value is not set
+     *
+     * @throws \InvalidArgumentException
+     */
+    public function validateConfiguration()
```
benjaminkott: Missusing the validation function to set meaningful defaults, should be done correctly after making the processor standalone.

Classes/DataProcessing/LanguageMenuProcessor.php
```diff
@@ -0,0 +1,333 @@
...
+
+    /**
+     * Add more language properties
+     */
+    public function prepareLevelLanguageConfiguration()
+    {
+        parent::prepareLevelLanguageConfiguration();
+
+        if ($this->menuConfig['special'] === 'language') {
+            unset($this->menuLevelConfig['stdWrap.']['cObject.']['10']);
```
benjaminkott: This and following will be obsolete after making it standalone.

Classes/DataProcessing/LanguageMenuProcessor.php
```diff
@@ -0,0 +1,333 @@
...
+    }
+
+    /**
+     * Gets the language title for the language id submitted in $conf in JSON format
+     *
+     * @param string Empty string (no content to process)
+     * @param array TypoScript configuration
+     * @return string JSON encoded data
+     */
+    public function getTitleAsJson($content, $conf)
```
benjaminkott: Can we combine this and the following into a universal function?

gilbertsoft:
@benjaminkott thanks for your comments will check and integrate them later. i just pushed my last changes but not tested so far.

Classes/DataProcessing/LanguageMenuProcessor.php
gilbertsoft: but that's what this code does!? i load the language ids from sys_language if special.value is not set or do i miss something? the only needed change i see here is to use a user object instead of text to be able to difference between with or without sites extension for v9...

Classes/DataProcessing/LanguageMenuProcessor.php
gilbertsoft: to clarify my idea about the titles also told to benni and he thinks also about this, but no feedback so far: the title is used primary for the be, so in editors language or english. if this differs from the fe user language what's normally the case the editor can set the nav_title with the destination language of the fe user. so it should be exactly what you wrote. have a look at this https://github.com/gilbertsoft/typo3-sites/blob/fdf7713f871637752dc1816aa2c0d0b307fe9676/Configuration/SiteConfigurationTCA/sys_site_language.php

Classes/DataProcessing/LanguageMenuProcessor.php
gilbertsoft: yeah, was the only way with the menuprocessor, will become very different now...

Classes/DataProcessing/LanguageMenuProcessor.php
gilbertsoft: yes, will take a look. but what i don't like is to use $conf for that and don't see another way now. but will check again later...

Classes/DataProcessing/LanguageMenuProcessor.php
gilbertsoft: Rewrite is complete

Classes/DataProcessing/LanguageMenuProcessor.php
gilbertsoft: Changed, now a optional parameter languages is supported

Classes/DataProcessing/LanguageMenuProcessor.php
gilbertsoft: done

Classes/DataProcessing/LanguageMenuProcessor.php
gilbertsoft: ah, now i understood. change is pending ;)

Classes/DataProcessing/LanguageMenuProcessor.php
gilbertsoft: done

Classes/DataProcessing/LanguageMenuProcessor.php
gilbertsoft: done

Classes/DataProcessing/LanguageMenuProcessor.php
gilbertsoft: done, i now use a second parameter `field` to define the desired field in ts

gilbertsoft:
@benjaminkott @cedricziel Major is done, some minor changes e.g. sorting and documentation must be done. Let the review begin...

Classes/DataProcessing/LanguageMenuProcessor.php
```diff
@@ -0,0 +1,494 @@
...
+
+    /**
+     * Process additional data processors
+     *
+     * @param array $page
+     * @param array $processorConfiguration
+     */
+    protected function processAdditionalDataProcessors($page, $processorConfiguration)
+    {
+        if (is_array($page['children'])) {
```
benjaminkott: Children will not be available here, can be removed.

Classes/DataProcessing/LanguageMenuProcessor.php
```diff
@@ -0,0 +1,494 @@
...
+
+        // Return processed data
+        $processedData[$this->menuTargetVariableName] = $processedMenu;
+        return $processedData;
+    }
+
+    /**
+     * Process additional data processors
+     *
+     * @param array $page
```
benjaminkott: `$page` should be `$language`

Classes/DataProcessing/LanguageMenuProcessor.php
```diff
@@ -0,0 +1,494 @@
...
+                    'stdWrap.' => [
+                        'wrap' => '"title":|'
+                    ]
+                ],
+                '21' => 'TEXT',
+                '21.' => [
+                    'value' => self::LINK_PLACEHOLDER,
+                    'wrap' => ',"link":|',
+                ],
+                '22' => 'TEXT',
```
benjaminkott: do we really need a target here?

Configuration/TypoScript/Compatibility/constants.typoscript
```diff
@@ -0,0 +1,21 @@
+#####################################
+### PAGE compatibility for T3 < 9 ###
+#####################################
+page {
+    theme {
```
benjaminkott: i think this should be moved one level up to `page.language` or even move to `config.language` because this is a global setting and not "theme" specific.

Configuration/TypoScript/constants.typoscript
```diff
@@ -108,8 +108,8 @@ page {
         language {
             # cat=bootstrap package: basic/168/language_enable; type=boolean; label=LanguageSection: Enable to display the LanguageSection at bottom
             enable = 1
-            # cat=bootstrap package: basic/168/language_languageValue; type=string; label=Language Navigation Values: Comma separated list of language uids
-            languageValue = 0,1,2
+            # cat=bootstrap package: basic/168/language_languageValue; type=string; label=Language Navigation Values: Comma separated list of language uids. Leave empty to load all enabled system languages
+            languageValue =
```
benjaminkott: i think we should make this more speaking, instead of providing an empty value we shoud really make clear that this is auto generated `language.languageValue = auto`

Classes/DataProcessing/LanguageMenuProcessor.php
```diff
@@ -0,0 +1,494 @@
...
+    protected function processAdditionalDataProcessors($page, $processorConfiguration)
+    {
+        if (is_array($page['children'])) {
+            foreach ($page['children'] as $key => $item) {
+                $page['children'][$key] = $this->processAdditionalDataProcessors($item, $processorConfiguration);
+            }
+        }
+        /** @var ContentObjectRenderer $recordContentObjectRenderer */
+        $recordContentObjectRenderer = GeneralUtility::makeInstance(ContentObjectRenderer::class);
+        $recordContentObjectRenderer->start($page['data'], 'pages');
```
benjaminkott: Do we really have a usecase for additional data processors here?

Classes/DataProcessing/LanguageMenuProcessor.php
```diff
@@ -0,0 +1,494 @@
...
+
+        // Build Configuration
+        $this->prepareConfiguration();
+        $this->buildConfiguration();
+
+        // Process Configuration
+        $menuContentObject = $cObj->getContentObject('HMENU');
+        $renderedMenu = $menuContentObject->render($this->menuConfig);
+        if (!$renderedMenu) {
+            return $processedData;
```
benjaminkott: where does `$processedData` come from?

Classes/DataProcessing/LanguageMenuProcessor.php
```diff
@@ -0,0 +1,494 @@
...
+     * @return array the processed data as key/value store
+     */
+    public function process(ContentObjectRenderer $cObj, array $contentObjectConfiguration, array $processorConfiguration, array $processedData)
+    {
+        $this->cObj = $cObj;
+        $this->processorConfiguration = $processorConfiguration;
+
+        // Get Configuration
+        $this->menuTargetVariableName = $this->getConfigurationValue('as');
+        $this->menuAlternativeSortingField = $this->getConfigurationValue('alternativeSortingField');
```
benjaminkott: already mentioned before do we need an alternative sorting field? does this work as exptected so we can sort the language records? i think a sorting could make sense but should be applied on the resultset and we do a simple array sorting here.

Classes/DataProcessing/LanguageMenuProcessor.php
```diff
@@ -0,0 +1,494 @@
...
+     */
+    protected $processorConfiguration;
+
+    /**
+     * Allowed configuration keys for menu generation, other keys
+     * will throw an exception to prevent configuration errors.
+     *
+     * @var array
+     */
+    public $allowedConfigurationKeys = [
```
benjaminkott: i think we can drop some more values here, since we need to ensure that this also works without `HMENU` at some point. having too much options that we would need to rebuild would make it for the core extremele hard to make changes here because its all public api.

```
cache_period
minItems
minItems.
maxItems
maxItems.
begin
begin.
alternativeSortingField
alternativeSortingField.
excludeUidList
excludeUidList.
excludeDoktypes
alwaysActivePIDlist
alwaysActivePIDlist.
protectLvar
addQueryString
addQueryString.
```

Classes/DataProcessing/LanguageMenuProcessor.php
```diff
@@ -0,0 +1,494 @@
...
+        'alwaysActivePIDlist.',
+        'protectLvar',
+        'addQueryString',
+        'addQueryString.',
+        'if',
+        'if.',
+        'languages',
+        'languages.',
+        'as',
+        'dataProcessing',
```
benjaminkott: if we really want to provide support for additional `dataProcessors` we need to define on what dataset we are working.

Classes/DataProcessing/LanguageMenuProcessor.php
```diff
@@ -0,0 +1,494 @@
...
+use TYPO3\CMS\Frontend\ContentObject\DataProcessorInterface;
+
+/**
+ * This menu processor utilizes HMENU to generate a json encoded menu
+ * string that will be decoded again and assigned to FLUIDTEMPLATE as
+ * variable. Additional DataProcessing is supported and will be applied
+ * to each record.
+ *
+ * Options:
+ * as - The variable to be used within the result
```
benjaminkott: Add missing options, like `languages` and `dataProcessing` if this is found to be useful.

Classes/DataProcessing/LanguageMenuProcessor.php
```diff
@@ -0,0 +1,494 @@
...
+ *     10 {
+ *        references.fieldName = media
+ *     }
+ *   }
+ * }
+ */
+class LanguageMenuProcessor implements DataProcessorInterface
+{
+    const LINK_PLACEHOLDER = '###LINKPLACEHOLDER###';
+    const TARGET_PLACEHOLDER = '###TARGETPLACEHOLDER###';
```
benjaminkott: Do we really need a target?

Classes/DataProcessing/LanguageMenuProcessor.php
```diff
@@ -0,0 +1,494 @@
...
+                    ],
+                    'wrap' => ',"languageUid":"|"'
+                ],
+                '70' => 'USER',
+                '70.' => [
+                    'userFunc' => 'TYPO3\CMS\Frontend\DataProcessing\LanguageMenuProcessor->getFieldAsJson',
+                    'language.' => [
+                        'data' => 'register:languageUid'
+                    ],
+                    'field' => 'nav_title',
```
benjaminkott: still unsure about the nav_title thing, in case of we need it as separeate field - because would not like to change templates and write fallbacks when nav_title is not set in my template - i would like to see a field i can rely on that this one is filled with either nav_title or title.

Classes/DataProcessing/LanguageMenuProcessor.php
```diff
@@ -0,0 +1,494 @@
...
+     * @var array
+     */
+    public $menuDefaults = [
+        'as' => 'languagemenu'
+    ];
+
+    /**
+     * @var string
+     */
+    protected $menuAlternativeSortingField;
```
benjaminkott: are we able to sort this? is it passed to `sys_language` for sorting?

Classes/Hooks/Frontend/TSGenerator.php
```diff
@@ -0,0 +1,88 @@
...
+    const SYS_LANGUAGE_UID_PLACEHOLDER = '###SYSLANGUAGEUIDPLACEHOLDER###';
+    const LANGUAGE_PLACEHOLDER = '###LANGUAGEPLACEHOLDER###';
+    const LOCALE_PLACEHOLDER = '###LOCALEPLACEHOLDER###';
+    const HREF_LANG_PLACEHOLDER = '###HREFLANGPLACEHOLDER###';
+    const DIRECTION_PLACEHOLDER = '###DIRECTIONPLACEHOLDER###';
+
+    /**
+     * @var string
+     */
+    protected $tempDirectory = 'typo3temp/assets/bootstrappackage/';
```
benjaminkott: still unsure about the location here. the assets folder is not protected because we will have public data here, the language configuration is private stuff and should never be available anywhere publicly.

Classes/DataProcessing/LanguageMenuProcessor.php
gilbertsoft: as far as i see the processor makes currently no sense. once HMENU or i more prefer a new content object in the core because we are creating flat menus here and nothing hierarchical has changed maybe we can use it to load e.g. flags.

Classes/DataProcessing/LanguageMenuProcessor.php
gilbertsoft: done

Classes/DataProcessing/LanguageMenuProcessor.php
gilbertsoft: for me it's not really useful but what if hmenu delivers it?

Configuration/TypoScript/constants.typoscript
gilbertsoft: yeah, that's a nice idea and done

Configuration/TypoScript/Compatibility/constants.typoscript
gilbertsoft: mmh, currently i don't think it will be used for the core because there we always should take the defaults from sites ext. but moving up makes sense.

Classes/DataProcessing/LanguageMenuProcessor.php
gilbertsoft: no, is't and see comment above

Classes/DataProcessing/LanguageMenuProcessor.php
gilbertsoft: function parameter

Classes/DataProcessing/LanguageMenuProcessor.php
gilbertsoft: yeah, sorting is pending as written. will do this later...

Classes/DataProcessing/LanguageMenuProcessor.php
gilbertsoft: agree, but i wasn't sure which one's i can drop, please make a suggestion

Classes/DataProcessing/LanguageMenuProcessor.php
gilbertsoft: see above

Classes/DataProcessing/LanguageMenuProcessor.php
gilbertsoft: see above

Classes/DataProcessing/LanguageMenuProcessor.php
gilbertsoft: i already integrated a fallback see https://github.com/gilbertsoft/bootstrap_package/blob/c3e0dd3297ac60ec6c4bdfa49d100640c02cf156/Classes/Utility/LanguageUtility.php#L84. i think having two fields is more flexible and with the fallback it should run out of the box

Classes/DataProcessing/LanguageMenuProcessor.php
gilbertsoft: see above, sorting is pending

Classes/Hooks/Frontend/TSGenerator.php
gilbertsoft: any suggestion?

gilbertsoft:
@benjaminkott push is done, please have a look at my comments...

Classes/Hooks/Frontend/TSGenerator.php
gilbertsoft: what about typo3temp/var/Cache/Code/bootstrappackage/?

Classes/DataProcessing/LanguageMenuProcessor.php
benjaminkott: we can just drop it, it hmenu would deliver it - the current page would be set to blank for exmaple, that would result in switching the language would open a new tab - i think we can savely drop it.

Classes/DataProcessing/LanguageMenuProcessor.php
benjaminkott: i think we can drop this.

Classes/DataProcessing/LanguageMenuProcessor.php
benjaminkott: well, the stuff above it the suggestion ;)

Classes/DataProcessing/LanguageMenuProcessor.php
benjaminkott: i would drop additional processors for now - we can add them as feature at a later point agian.

Classes/DataProcessing/LanguageMenuProcessor.php
benjaminkott: ok, cool - then everything is fine for me.

Classes/DataProcessing/LanguageMenuProcessor.php
benjaminkott: kk

Classes/DataProcessing/LanguageMenuProcessor.php
benjaminkott: kk

Classes/DataProcessing/LanguageMenuProcessor.php
benjaminkott: ok, missed that - thx for clarification

Classes/DataProcessing/LanguageMenuProcessor.php
benjaminkott: thx!

Classes/DataProcessing/LanguageMenuProcessor.php
benjaminkott: flags could now also be loaded easily via css, we can prepare selectors for that in the markup - i would drop as mentioned before additional data processors.

Classes/DataProcessing/LanguageMenuProcessor.php
benjaminkott: ok, lets drop them.

Classes/Hooks/Frontend/TSGenerator.php
benjaminkott: not sure about that one, but i know that there will be changes in the future to that path. see https://github.com/TYPO3/TYPO3.CMS/commit/6361660f70e6ee0398e5b68ae472633415c77103

Configuration/TypoScript/Compatibility/constants.typoscript
benjaminkott: talked to @psychomieze on sunday about this, current state is that sites will be optional for the beginning. so we could also make this available for cms9 when sites are not used.

Classes/DataProcessing/LanguageMenuProcessor.php
gilbertsoft: hey, that's from your code ;)

Configuration/TypoScript/constants.typoscript
benjaminkott: thx!

Classes/DataProcessing/LanguageMenuProcessor.php
gilbertsoft: done now

Classes/DataProcessing/LanguageMenuProcessor.php
gilbertsoft: done

Classes/DataProcessing/LanguageMenuProcessor.php
gilbertsoft: oops

benjaminkott:
@gilbertsoft thx for clarification added comments.

Classes/DataProcessing/LanguageMenuProcessor.php
gilbertsoft: done, but `alternativeSortingField` because it's supported now

Classes/DataProcessing/LanguageMenuProcessor.php
gilbertsoft: done

Classes/DataProcessing/LanguageMenuProcessor.php
gilbertsoft: please close this change request

Classes/DataProcessing/LanguageMenuProcessor.php
gilbertsoft: please close change request

Classes/DataProcessing/LanguageMenuProcessor.php
gilbertsoft: please close change request

Classes/DataProcessing/LanguageMenuProcessor.php
gilbertsoft: please close change request

Classes/DataProcessing/LanguageMenuProcessor.php
gilbertsoft: please close change request

gilbertsoft:
@benjaminkott almost everything is done now. i'm not sure about the line https://github.com/gilbertsoft/bootstrap_package/blob/ead5435d5baed9b66adbe0fb678dfdf13c09df47/Classes/DataProcessing/LanguageMenuProcessor.php#L406, what is this doing? can i remove it too? some `todo` is still open, you can search the code for it if interessted

Classes/Hooks/Frontend/TSGenerator.php
gilbertsoft: as far as i see this should not break my impl (which i realy don't like but i din't find an other way so far) other place coud be some where in ext:bootstrap_package/Resources/Private, also not nice but as far the file will be recreated should not make troubles...

gilbertsoft:
@benjaminkott @cedricziel everything is done so far and can't longer find any problems. only pending thing is test with sites, but didn't submit the required stuff to https://review.typo3.org/#/c/56505/ for now and not tested sites so far. will prepare this the next days...

gilbertsoft:
Tests have be done with TYPO3 master and the Sites module

gilbertsoft:
Sites are now optional and if not configured fallback to prior sites handling

Classes/DataProcessing/LanguageMenuProcessor.php
```diff
@@ -317,42 +300,29 @@ public function validateConfiguration()
             }
         }
         if (!empty($invalidArguments)) {
-            throw new \InvalidArgumentException('LanguageMenuProcessor Configuration contains invalid Arguments: ' . implode(', ', $invalidArguments), 1478806566);
+            throw new \InvalidArgumentException('LanguageMenuProcessor configuration contains invalid arguments: ' . implode(', ', $invalidArguments), 1522959188);
```
gilbertsoft: error code should be regenerated to not to be equal with the core

Classes/DataProcessing/LanguageMenuProcessor.php
```diff
@@ -279,36 +265,33 @@ public function getFieldAsJson($content, $conf)
...
+
+        // Throws InvalidArgumentException in case language is not found which is fine
+        $language = $site->getLanguageById((int)$conf['language'])->toArray();
 
-        $row = LanguageUtility::getLanguageRow($conf['page'], $conf['language']);
-        if (isset($row[$conf['field']])) {
-            $result = $this->jsonEncode($row[$conf['field']]);
+        // Check field for return exists
+        if ($language !== null && !isset($language[$conf['field']])) {
+            throw new \InvalidArgumentException('Invalid value \'' . $conf['field'] . '\' for argument \'field\' supplied.', 1524063160);
```
gilbertsoft: error code should be regenerated

Classes/DataProcessing/LanguageMenuProcessor.php
```diff
@@ -279,36 +265,33 @@ public function getFieldAsJson($content, $conf)
...
         if ($conf['language'] === '') {
-            throw new \InvalidArgumentException('Argument \'language\' must be supplied.', 1522795243);
+            throw new \InvalidArgumentException('Argument \'language\' must be supplied.', 1522959186);
         }
         if ($conf['field'] === '') {
-            throw new \InvalidArgumentException('Argument \'field\' must be supplied.', 1522795274);
-        }
-        if ($conf['page'] === '') {
-            throw new \InvalidArgumentException('Argument \'page\' must be supplied.', 1523106560);
+            throw new \InvalidArgumentException('Argument \'field\' must be supplied.', 1522959187);
```
gilbertsoft: error code should be regenerated

Classes/DataProcessing/LanguageMenuProcessor.php
```diff
@@ -20,23 +21,20 @@
  * decoded again and assigned to FLUIDTEMPLATE as variable.
  *
  * Options:
- * alternativeSortingField -
```
gilbertsoft: why do you remove the sorting support?

Classes/Hooks/Frontend/TypoScriptLanguageHook.php
```diff
@@ -45,30 +69,31 @@ class TypoScriptLanguageHook
...
-     * @param TemplateService $tmpl
+     * @param TemplateService $templateService
      */
-    public function addLanguageConditions(&$params, &$tmpl)
+    public function addLanguageConditions(&$params, &$templateService)
     {
-        $tmpl->config[] = $this->createLanguageConditions(method_exists($tmpl, 'getRootId') ? $tmpl->getRootId() : 0);
+        $rootPage = method_exists($templateService, 'getRootId') ? $templateService->getRootId() : $params['rootLine'][0]['uid'];
+        ExtensionManagementUtility::addTypoScriptConstants(implode(LF, $this->constants));
+        ExtensionManagementUtility::addTypoScriptSetup($this->createLanguageConditions($rootPage));
```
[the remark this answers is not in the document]
benjaminkott: problem with your solution was that it was not correctly applied to all areas like typoscript inspector etc. while you are adding directly to the configuration the scripts are adding it to the default typoscript configuration that means it can be overridden more easily, is processed correctly and shows up in constants editor etc.

Classes/DataProcessing/LanguageMenuProcessor.php
```diff
@@ -279,36 +265,33 @@ public function getFieldAsJson($content, $conf)
...
-            throw new \InvalidArgumentException('Argument \'field\' must be supplied.', 1522795274);
-        }
-        if ($conf['page'] === '') {
-            throw new \InvalidArgumentException('Argument \'page\' must be supplied.', 1523106560);
+            throw new \InvalidArgumentException('Argument \'field\' must be supplied.', 1522959187);
         }
 
-        $result = '';
+        // Get and check current site
+        $site = $this->getCurrentSite();
```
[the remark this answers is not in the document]
benjaminkott: site has been removed completely, only support for 8.7 and master (e.g. 9.3)

Classes/DataProcessing/LanguageMenuProcessor.php
benjaminkott: should be streamlined

Classes/DataProcessing/LanguageMenuProcessor.php
benjaminkott: should be streamlined

Classes/DataProcessing/LanguageMenuProcessor.php
benjaminkott: should be fine

Classes/DataProcessing/LanguageMenuProcessor.php
benjaminkott: sorting has been removed since master and 9.3 implementation does currently not support this - to avoid confusions when people are upgrading a feature could be missing. the goal is to polyfill the 9.3 behavior,

Classes/Hooks/Frontend/TypoScriptLanguageHook.php
```diff
@@ -45,30 +69,31 @@ class TypoScriptLanguageHook
...
      * @param array $params
-     * @param TemplateService $tmpl
+     * @param TemplateService $templateService
      */
-    public function addLanguageConditions(&$params, &$tmpl)
+    public function addLanguageConditions(&$params, &$templateService)
     {
-        $tmpl->config[] = $this->createLanguageConditions(method_exists($tmpl, 'getRootId') ? $tmpl->getRootId() : 0);
+        $rootPage = method_exists($templateService, 'getRootId') ? $templateService->getRootId() : $params['rootLine'][0]['uid'];
+        ExtensionManagementUtility::addTypoScriptConstants(implode(LF, $this->constants));
```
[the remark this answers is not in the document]
benjaminkott: should be fine

Classes/DataProcessing/LanguageMenuProcessor.php
gilbertsoft: that's fine, we could readd it if sorting of core is done if desired.

PR #526, opened 2018-06-03
add event for OnPopupClose in cookieconsent

buechlerpro:
### Description

A user might apply additional margin to the topmost or bottommost container to avoid the cookiedialog to overlap tha page content. When closing the cookiedialog these margins can be removed.


benjaminkott:
@buechlerpro thx a lot for your contribution, ive also added the open event and added the missing compiled files

PR #546, opened 2018-08-15
[BUGFIX] Get rid of warning for count() thrown in less compiler with PHP 7.2

ute-arbeit:
As of PHP 7.2 a warning is thrown when calling count() with a parameter that is a scalar, null, or an object that doesn't implement Countable.

Fix by checking whether the argument implements the Countable interface.

* [x ] Changes have been tested on TYPO3 8.7 LTS


benjaminkott:
@ute-arbeit thx a lot for the patch!

PR #551, opened 2018-09-24
[TASK] Code cleanups

NeoBlack:


composer.json
```diff
@@ -27,8 +27,19 @@
...
+    "typo3/cms-backend": "^8.7.0 || ^9.0",
     "typo3/cms-core": "^8.7.0 || ^9.0",
-    "typo3/cms-rte-ckeditor": "^8.7.0 || ^9.0"
+    "typo3/cms-extbase": "^8.7.0 || ^9.0",
+    "typo3/cms-extensionmanager": "^8.7.0 || ^9.0",
+    "typo3/cms-fluid": "^8.7.0 || ^9.0",
+    "typo3/cms-frontend": "^8.7.0 || ^9.0",
+    "typo3/cms-install": "^8.7.0 || ^9.0",
+    "typo3/cms-rte-ckeditor": "^8.7.0 || ^9.0",
+    "typo3fluid/fluid": "^2.3 || ^2.5.2"
```
benjaminkott: i know you do not like this but i would not like to have the direct dependency to fluid here, we can work with the version that typo3/cms-fluid defined - this will make too much problems in the future when core updates to a new major version i need to do a release to catch this one.

NeoBlack: I don't see a problem here, but it's your package... so if you don't want this it is ok for me too.

composer.json
```diff
@@ -27,7 +27,17 @@
   },
   "require": {
     "php": ">=7.0.0",
+    "ext-dom": "*",
+    "ext-libxml": "*",
+    "ext-simplexml": "*",
+    "leafo/scssphp": "^0.7.4",
```
benjaminkott: please remove we currently only use the bundled version - composer integration of compiling libs need a different handling

PR #585, opened 2018-11-06
[BUGFIX] Parse SCSS files correctly when using symlinked paths

ohader:
+ project/public/typo3conf/ext/bootstrap_package
+ project/packages/some_site_package
  + having `@import "../../../../../bootstrap_package/Resources/Public/Scss/Theme/theme"` in source file
  + (generated using https://www.sitepackagebuilder.com/)

### DIFF
--- a/Classes/Parser/ScssParser.php
+++ b/Classes/Parser/ScssParser.php
@@ -14,6 +14,7 @@ use Leafo\ScssPhp\Formatter\Crunched;
 use Leafo\ScssPhp\Version;
 use TYPO3\CMS\Core\Utility\ExtensionManagementUtility;
 use TYPO3\CMS\Core\Utility\GeneralUtility;
+use TYPO3\CMS\Core\Utility\PathUtility;
 
 /**
  * ScssParser
@@ -83,13 +84,19 @@ class ScssParser extends AbstractParser
                 'sourceMapBasepath' => '<PATH DOES NOT EXIST BUT SUPRESSES WARNINGS>'
             ]);
         }
-        $css = $scss->compile('@import "' . $file . '"');
+
+        $css = $scss->compile($this->resolveCode($file));
 
         $absoluteFilename = GeneralUtility::getFileAbsFileName($file);
         $relativePath = $settings['cache']['tempDirectoryRelativeToRoot'] . dirname(substr($absoluteFilename, strlen($this->getPathSite()))) . '/';
         $search = '%url\s*\(\s*[\\\'"]?(?!(((?:https?:)?\/\/)|(?:data:?:)))([^\\\'")]+)[\\\'"]?\s*\)%';
         $replace = 'url("' . $relativePath . '$3")';
         $css = preg_replace($search, $replace, $css);
+
+        $parsedFiles = $scss->getParsedFiles();
+        if (!isset($parsedFiles[$absoluteFilename])) {
+            $parsedFiles[$absoluteFilename] = filemtime($absoluteFilename);
+        }
 
         return [
             'css' => $css,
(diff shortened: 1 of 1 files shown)


ohader:
Can you please provide additional information about the error message, stack trace and the scenario you've been using? Thx in advance!

Classes/Parser/ScssParser.php
```diff
@@ -83,25 +84,88 @@ protected function parseFile($file, $settings)
...
+        $visualFilePath = GeneralUtility::getFileAbsFileName($file);
+        $realFilePath = realpath($visualFilePath);
+        if ($visualFilePath === $realFilePath) {
+            return '@import "' . $file . '"';
+        }
+
+        $visualDirectory = dirname($visualFilePath);
+        $realDirectory = dirname($realFilePath);
+        $code = file_get_contents($visualFilePath);
+        if (preg_match_all('#@import\s+"([^"]+)"#', $code, $matches)) {
```
helhum: If i understand it correctly, the issues can only occur in one file, which is the entry point file for parsing, which can reference other scss files on other extensions.

Two suggestions how this can be resolved cleanly:

1. Support `EXT:extension_key` syntax and replace all occurrences with absolute paths (similar to the approach here, but a bit more predictable).

2. Always use absolute fully qualified paths with a stream wrapper URI like that: `typo3://ext_key/Resources/...` and register a stream wrapper that can resolve these to absolute paths. Have not checked that, but the SCSS parser should be able to handle such paths. Benefit: no parsing needs to be done in the source files and any file can reference other files from extensions that way easily.

ohader: > If i understand it correctly, the issues can only occur in one file, which is the entry point file for parsing, which can reference other scss files on other extensions.

Basically it can happen in the whole `@import` chain and every time paths to other components are defined using path-traversals (`../../../whatever`).

on 1) okay, needs additional adjustments in generated code (e.g. created from https://www.sitepackagebuilder.com/ - that was my use-case)
* however this (still) only works on the caller side, since 3rd party library does not know anything about `EXT:...`
* (it feels like the "crash" @benjaminkott referred to is caused by pre-parsing and substituting paths during runtime - the patch does not have memory management nor detection for infinite loops)

on 2) the cleanest solution but also the one that cannot be used "right now"
* requires adjustments on existing (generated) code (see previous remark)
* `typo3://` (or any other stream wrapper) should be defined carefully in order to avoid confusion - this would be something for TYPO3 core and not only for the Bootstrap Package parsing sub component

ohader:
Anyway... looks like another year I'm using

```
"extra": {
        "patches": {
            "bk2k/bootstrap-package": {
                "Processes symlinked SCSS sources":
"https://patch-diff.githubusercontent.com/raw/benjaminkott/bootstrap_package/pull/585.patch" }
} }```

gilbertsoft:
@ohader Had the same issue and an other concerning the `@import` directive which I tried to patch but they didn't accept it see https://github.com/scssphp/scssphp/pull/57. So I will try to find a solution here and hopefully you don't have to patch for an other year...

Classes/Parser/ScssParser.php
helhum: > the cleanest solution

Glad we have an agreement here 👍

> but also the one that cannot be used "right now"

True. However, when we agree and join forces, it could be used in a very near future (closer than "another year").

> typo3:// (or any other stream wrapper) should be defined carefully in order to avoid confusion

I agree.

> this would be something for TYPO3 core and not only for the Bootstrap Package parsing sub component

Well, `typo3://` was just an example. Could be anything and the wrapper could be registered only for parsing and removed after parsing to lower conflict possibilities. E.g. `scssfiles://foo/main.scss` which resolves to `typo3conf/ext/foo/Resources/Private/Styles/Scss/main.scss`, or even cleaner, introduce a registry for path mappings.

I'd contribute the code for this package here, but only if it is accepted as solution to avoid working for the bin.

Issue #608, opened 2018-12-02
The classes .frame-ruler-before and .frame-ruler-after are missing pseudo classes used in fluid_styled_content


Outdoorsman:
# Bug Report

## Prerequisites

* [x] Can you reproduce the problem all the way from TYPO3 v8.0 to v9.5
* [x] Did you [perform a cursory search](https://github.com/issues?q=is%3Aissue+user%3Abenjaminkott+repo%3Abenjaminkott%2Fbootstrap_package) to see if your bug or enhancement is already reported?

## Description

When the class `.frame-ruler-after` is used it causes two borders instead of one among other things. This is also true for `.frame-ruler-before` class.

The reason? It appears that `bootstrap_package` is attempting to overriding the `.frame-ruler-before:before` and `frame-ruler-after:after` that are in the `fluid_styled_content` extension, but it does not include the pseudo class so instead of overriding it's causing duplicate CSS. Here's a sample of what what using the `frame-ruler-after` looks like on a real website. This issue goes back a ways so might be good to release fixes for the other releases of `bootstrap_package` beside just the current version 10???

![2018-12-01_19-47-55](https://user-images.githubusercontent.com/3075533/49335489-0197cc80-f5a3-11e8-89dd-dbc5a643de75.png)

## Steps to Reproduce

1. Make sure that `bootstrap_package` and `fluid_styled_content` are both installed an active.
2. Add a content element to the website and select _Ruler After_ or _Ruler Before_ from the dropdown in the _Appearances_ tab and save.

**Expected behavior:** [What you expected to happen]

There should a single border either above or below the content element based on your selection.

**Actual behavior:** [What actually happened]

There are two borders above or below based on your selection.

## Versions

You can get this information from executing `composer show` at the command line if your TYPO3 installation is composer based, otherwise the version number will be shown in the extension manager.

This conflict occurs for `bootstrap_package`:

- The file [Resources/Public/Less/Theme/frame.less](https://github.com/benjaminkott/bootstrap_package/blob/master/Resources/Public/Less/Theme/frame.less) in `bootstrap_package` v8.0.0 to v10.0.3
- The file [Resources/Public/Scss/Theme/_frame.scss](https://github.com/benjaminkott/bootstrap_package/blob/master/Resources/Public/Scss/Theme/_frame.scss) in `bootstrap_package` v10.0.0 to v10.0.3

When this file is present in `fluid_styled_content`:

[...]

PR #609, opened 2018-12-02
[BUGFIX] Add missing pseudo class to match fluid_styled_content

Outdoorsman:
Fixes #609.

See #609 for more details about what this does.


benjaminkott:
@Outdoorsman thx for the patch, but the current behavior is intended. Please have a look at the comment in https://github.com/benjaminkott/bootstrap_package/issues/608

PR #625, opened 2018-12-18
[BUGFIX] Typo: 'boostrap' should be 'bootstrap'

lolli42:


wtfred:
Glad to see i'm not the only one constantly making this typo :smile:

PR #633, opened 2019-01-19
Link on #top

typo3ua:
I am propose change content in the file 
`bootstrap_package/Resources/Private/Partials/Page/Structure/ScrollTop.html`

### From:
```
<a class="scroll-top" title="{f:translate(key: 'scrolltop', extensionName: 'bootstrap_package')}" href="#top">
   <span class="scroll-top-icon"></span>
</a>
```
- It is with TS property 'baseURL` does not work
- It is without TS property `baseURL` does work
### To: 
```
<f:link.page class="scroll-top" title="{f:translate(key: 'scrolltop', extensionName: 'bootstrap_package')}" section="top">
    <span class="scroll-top-icon"></span>
</f:link.page>
```
- It is with TS property `baseURL` does work
- It is without TS property `baseURL` does work

### Prerequisites

### Description

My proposed works well with TS property `baseURL` and without `baseURL` too
```
config.baseURL = https//www.domain.local/
```


benjaminkott:
This pull request seems to be also messed up, also I have to decline the proposed change. The usage of baseurl is highly discouraged since TYPO3 6.2 and you should use absRefPrefix instead. There is also no need to set the baseUrl or absRefPrefix anyways since TYPO3 is able to resolve it correctly without configuration. The usage of the link viewhelper will also introduce new problems like, params are not attached for example details actions of extensions.

PR #724, opened 2019-07-03
Fix `custom-select-indicator`

sgrossberndt:
Due to a wrong `!default` specification the `custom-select-indicator` in a `custom-select` was not displayed.
Removing the wrong specification fixes this.


benjaminkott:
@sgrossberndt thx a lot!

PR #750, opened 2019-09-21
[FEATURE] Renders shortcuts im main menu as active

buepro:
### Prerequisites

* Changes have been tested on TYPO3 v9.5 LTS
* Changes have been tested on TYPO3 dev-master
* Changes have been tested on PHP 7.2.x

### Description

In some cases the menu entry linking to the main page is a shortcut
pointing to its parent page. When the parent page is shown the
shortcut page isn't shown as active. With this branch the shortcut
is shown as active.

### DIFF
--- a/Resources/Private/Partials/Page/Navigation/Main.html
+++ b/Resources/Private/Partials/Page/Navigation/Main.html
@@ -24,7 +24,7 @@
             </button>
             <nav id="mainnavigation" class="collapse navbar-collapse">
                 <f:render partial="DropIn/Navigation/MainBefore" arguments="{_all}" />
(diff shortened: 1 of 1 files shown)


gilbertsoft:
Hey @buepro 

Thanks for your proposition on this issue (https://forge.typo3.org/issues/85138).

Fixing this in the fluid template is the wrong place to do it, this will have side effects on caching e.g. This should be fixed at the core but looks easier as it really is because the HMENU CE is a really complex part of the core.

Next possibility would be a fix in the MenuProcessor, this is also mentioned in the core issue.

I will try to dig again into this issue and provide a core fix for version 10. For older version we will discuss about a fix in the MenuProcessor here in Bootstrap Package.

I've created now an issue here #829 and will close your PR, but thanks again for working on this!

PR #773, opened 2019-11-06
[DOCS] Add TYPO3 Render Documentation container

gilbertsoft:


benjaminkott:
I know you love this, but I would prefer to add docs on how to use the rendering container from the documentation team.

gilbertsoft:
I don't get your comment fully, the documentation of the docs team renderer is very detailed but is not very user friendly imho. This patch uses the container of the docs team and makes some minor changes to it to get it running with DDEV (I have also a simpler version without changing the container but there is no healthcheck then). The original container is meant to die, you call it, docu is rendered and the container will be terminated. DDEV does not support dying containers so the simple usage with DDEV needs some changes. And this is still WIP, I requested some changes for DDEV to be able to finish this one. Imo it's a nice progress to simply call `ddev t3rd ...` and your docu is rendered the TYPO3 docs team way instead of doing this crazy and not plattform independant source thing see https://docs.typo3.org/m/typo3/docs-how-to-document/master/en-us/RenderingDocs/Quickstart.html#commands-to-render-documentation

gilbertsoft:
drud/ddev#1957

gilbertsoft:
Closing this one in favor of #901

PR #783, opened 2019-11-16
[TASK] Update ddev to 1.12.0-alpha2

gilbertsoft:
Or should we wait until the final release?


benjaminkott:
Would wait for the final release.

PR #786, opened 2019-11-17
[BUGFIX] Escape special chars in bootstrap.smoothscroll.js

### LINKED ISSUE #759: bootstrap.smoothscroll.js throwing error if hash ID contains a leading slash from URL slug (Fix provided)
# Bug Report

For a OnePage layout I'm generating sections for each page, using the page URL slug as ID, like http://mysite.net/#/services.
Unfortunately the script `bootstrap.smoothscroll.js` (or jQuery to be precise) can't handle hash values starting with a leading slash (`Error: Syntax error, unrecognized expression: #/services`).

# Bug Fix

In file `bootstrap.smoothscroll.js`, line 11 we need to escape special chars inside the hash:

```
OLD: var $target = $(this.hash);
NEW: var $target = $(this.hash.replace( /(:|\.|\[|\]|,|=|\/)/g, '\\$1'));
```

Have tested this with bootstrap_package v11.0.1 in TYPO3 v9.5.8 and its working again like expected.

gilbertsoft:
Fixes #759

@benjaminkott some notes on this patch:
- I had to update node-sass to support nodejs 12. Should we make this better in an own patch?
- modernizr changed too I guess because of the previous update. How to proceed here?
- and finally please check the changes in smoothscroll


benjaminkott:
@gilbertsoft would make it easier to follow the changelog if we upgrade node to v12 in a dedicated patch, also the built environment needs to be adapted.

gilbertsoft:
see #790

PR #787, opened 2019-11-17
[TASK] Add BackendLayouts to Web Info Module

### LINKED ISSUE #748: Get pagelayout for each page in menu
# Support

Is there a way to get the pagelayout (backendlayout & backend_layout_next_level) for each page in a menu. i only get the pagelayout form the current page, but not form the target pages of the menu entries.

gilbertsoft:
Fixes #748

### Prerequisites

### Description
Adds the backend layouts to the web info module

### Steps to Validate

1. Install Web Info Module (`typo3/cms-info`)
2. Open Web -> Info
3. In the Pagetree overview select the new item `Backend Layouts`


ext_localconf.php
```diff
@@ -32,6 +32,9 @@
     \TYPO3\CMS\Core\Utility\ExtensionManagementUtility::addPageTSConfig('<INCLUDE_TYPOSCRIPT: source="FILE:EXT:bootstrap_package/Configuration/TsConfig/Page/ContentElement/All.tsconfig">');
 }
 
+// BackendLayouts to the Web Info Module
+\TYPO3\CMS\Core\Utility\ExtensionManagementUtility::addPageTSConfig('<INCLUDE_TYPOSCRIPT: source="FILE:EXT:bootstrap_package/Configuration/TsConfig/Page/Mod/WebInfo/WebInfo.tsconfig">');
```
gilbertsoft: @benjaminkott I don't think we need the possibility to disable this like for the other PageTSConfig options

benjaminkott: I agree that we do not need this kind of option, but should we not add this to the core?

gilbertsoft: I'm not sure. For BP backend layouts are an important part and that must not be the case for every TYPO3 installation. @bmack what do you think?

bmack: feel free to create a FEATURE for core to add this.

gilbertsoft:
Patch for the core is done see https://review.typo3.org/c/Packages/TYPO3.CMS/+/62338

I will change this one to take effect for core versions which don't have this included by default.

ext_localconf.php
```diff
@@ -32,6 +32,14 @@
     \TYPO3\CMS\Core\Utility\ExtensionManagementUtility::addPageTSConfig('<INCLUDE_TYPOSCRIPT: source="FILE:EXT:bootstrap_package/Configuration/TsConfig/Page/ContentElement/All.tsconfig">');
 }
 
+// BackendLayouts to the Web Info Module
+if (stripos(
```
gilbertsoft: I don't like this but see not other solution...

PR #788, opened 2019-11-18
[TASK] Add masonry-layout as dependency

### LINKED ISSUE #768: masonry layout for gallery via backend option
# Feature Request

Right now you can create a gallery of images and each row has the same height. If you have a picture in portrait mode and several landscape images you will end up with white space under the landscape images.
see this: https://stackoverflow.com/questions/33512647/flexbox-stack-different-height-childs

It would be nice to have an option for a gallery you will get using the masonry library: https://masonry.desandro.com/

**Is your feature request related to a problem? Please describe.**
No

**Describe the solution you'd like**

1)
A solution could be this: http://thenewcode.com/844/Easy-Masonry-Layout-With-Flexbox

set `flex-direction:column` and a `height` to the gallery container.

2 problems I noticed:
- depending on the number ans sizes of images the `height` has to be variable. A fixed `height` does not work an every case.
- if you set the `flex-direction` to `column` the order of images is vertically and not chronological anymore

2)
another solution could be the use of css grid but that would need a fallback for IE
https://codepen.io/balazs_sziklai/pen/mOwoLg

gilbertsoft:
Fixes #768 

### Prerequisites

* [ ] Changes have been tested on TYPO3 v9.5 LTS
* [ ] Changes have been tested on PHP 7.2.x

### Description

This patch introduces the masonry layout see https://masonry.desandro.com/

### Steps to Validate

1. [First Step]
2. [Second Step]
3. [and so on...]


gilbertsoft:
@benjaminkott  I see here two possibilities to go ahead:

- Add a new CE for this feature with a new table for the images with the needed properties like col or row span. That's my favorite
- Extend the images in general with this properties which will make it possible to use collections too and the existing gallery

What's your opinion here?

gilbertsoft:
> What's your opinion here?

@medarob your feedback is ofc welcome too

medarob:
Hi,

from the user point of view I think a new option inside the gallery CE would be sufficient (if that is possible?).

It should be possible to switch the layouts without adding all images again. So if I create a standard bootstrap gallery then the user should have the option to switch to the masonry layout without creating a new CE and add all the images again. Maybe it's possible to change the `CType` and keep all the other data? Then a new CE would also work, I guess.

benjaminkott:
would prefer a simple css solution instead of a js lib for that. see:
https://w3bits.com/css-masonry/

gilbertsoft:
Another option, not checked so far https://www.smashingmagazine.com/native-css-masonry-layout-css-grid/

benjaminkott:
will not happen until https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_grid_layout/Masonry_layout

PR #790, opened 2019-11-18
[TASK] Update build dependencies

gilbertsoft:
This patch increases the build dependencies to support Node.js 12

- updated node-saas to ^4.12.0
- yarn install
- yarn build
- yarn css

I'm not sure if we need to increade Node.js to 12 here https://github.com/benjaminkott/bootstrap_package/blob/master/.github/workflows/ci.yml#L43

My patch should make sure Node.js 12 compatibility in the first place...


gilbertsoft:
...and for this we need to increase Node.js too. Got it now :D

PR #791, opened 2019-11-18
[BUGFIX] Format raw HTML by lib.parseFunc for link handling

### LINKED ISSUE #729: <link> and <a href="t3://page/uid=1"> not working in HTML CE
# Bug Report

`<link>` and `<a href="t3://page/uid=1">` tags are not parsed in HTML content elements because the `bootstrap_package/Resources/Private/Templates/ContentElements/Html.html` template contains `<f:format.raw>...</f:format.raw>` instead of:
```
    <f:format.html parseFuncTSPath="lib.linkParseFunc">{data.bodytext}</f:format.html>
```
The corresponding TS is:
```
lib.linkParseFunc {
    tags {
        link = TEXT
        link {
            stdWrap.current = 1
            stdWrap.typolink.extTarget = _blank
            stdWrap.typolink.target = {$cLinkTagTarget}
            stdWrap.typolink.parameter.data = parameters : allParams
        }
    }
}
```
This is only a partial solution because `<a href="t3://page/uid=1">` is still not parsed. Removing the `parseFuncTSPath` attribute will use the standard RTE parse functions but they destroy the HTML code and insert `<p>` tags all over.

gilbertsoft:
Fixes #729

### Prerequisites

### Description

This patch forces the parsing of raw HTML content by `lib.parseFunc` to ensure a correct link resolution and handling

### Steps to Validate

1. Create a HTML content element
2. Place some links with <a> or <link> tags
3. Verify the frontend output

### DIFF
--- a/Resources/Private/Templates/ContentElements/Html.html
+++ b/Resources/Private/Templates/ContentElements/Html.html
@@ -3,7 +3,7 @@
 <f:section name="Header" />
 <f:section name="Main">
 
-    <f:format.raw>{data.bodytext}</f:format.raw>
+    <f:format.html parseFuncTSPath="lib.parseFunc">{data.bodytext}</f:format.html>
 
 </f:section>
 </html>


gilbertsoft:
This change seems to be to risky to break existing installations. One solution would be to add an option per content element to activate the lib parser if needed.

benjaminkott:
This one is far to risky to approve, see html behavior in core - we really should handle it the same way here. https://github.com/TYPO3/TYPO3.CMS/blob/master/typo3/sysext/fluid_styled_content/Resources/Private/Templates/Html.html

Manawyrm:
I needed this behaviour and I just overwrote the Html.html template in my own extension with the version from this PR. 

We could just document this behaviour (maybe with a link in the Html.html for people who are looking for it?) and people who need it can overwrite the template themselves?

gilbertsoft:
> 
> 
> I needed this behaviour and I just overwrote the Html.html template in my own extension with the version from this PR.
> 
> We could just document this behaviour (maybe with a link in the Html.html for people who are looking for it?) and people who need it can overwrite the template themselves?

Good idea, maybe you can [add some words to the Wiki](https://github.com/benjaminkott/bootstrap_package/wiki/_new)?

PR #793, opened 2019-11-19
[TASK] Change build order

### LINKED ISSUE #789: bootstrappackageicon.min.css is missing after build
# Bug Report

## Description
After calling `yarn build` the file `bootstrappackageicon.min.css` is missing because a missing dependency to `cssmin` see https://github.com/benjaminkott/bootstrap_package/blob/master/Build/Gruntfile.js#L541. The task `webfont`must be moved before `css` or use the task `icon` instead which will call then `cssmin` a second time.

## Steps to Reproduce
1. `yarn build`
2. file `bootstrappackageicon.min.css` is deleted

gilbertsoft:
Fixes #789 

Steps to verify:

- delete Resources/Public/Fonts
- run `yarn build`
- verify all files are created and no one is missing


benjaminkott:
makes sense 👍

PR #796, opened 2019-11-20
[TASK] Automatically label dependency changes

gilbertsoft:
This patch activates the labeler action for pull request and defines the files containing dependencies to get the dependencies label added automatically.


gilbertsoft:
@benjaminkott Have just discovered something cool. Are there other dependency related files I forgot perhaps?

benjaminkott:
let's see how this works out 👍

PR #815, opened 2019-12-06
[BUGFIX] Install build tools

gilbertsoft:
This patch installs the build tools to the web container to enable the execution of yarn and grunt in the container.

Followup of #808


benjaminkott:
@gilbertsoft unnessesary, just call them through yarn or use the setup scripts

- `yarn grunt build`
- `yarn build`

PR #826, opened 2019-12-07
[!!!][TASK] Remove bootstrap3

### LINKED ISSUE #807: Remove Bootstrap 3 Support
Support for Boostrap version 3 will be removed for the next major release 12.0.0

gilbertsoft:
Closes #807


gilbertsoft:
@benjaminkott I'm not sure about this https://github.com/benjaminkott/bootstrap_package/blob/master/Build/templates/font.css#L32

Configuration/TCA/Overrides/101_sys_template.php
```diff
@@ -29,16 +29,6 @@
...
- * TypoScript: Framework
- * Include Bootstrap 3.x (LESS) Assets
- */
-\TYPO3\CMS\Core\Utility\ExtensionManagementUtility::addStaticFile(
-    'bootstrap_package',
-    'Configuration/TypoScript/Bootstrap3',
-    'Bootstrap Package: Bootstrap 3.x (LESS)'
-);
-
 /***************
```
benjaminkott: Should we keep the BP4 include? Not sure about BP5

benjaminkott:
> @benjaminkott I'm not sure about this https://github.com/benjaminkott/bootstrap_package/blob/master/Build/templates/font.css#L32

Also not sure about this - i think that was already in the template i adapted for the rendering. I guess it can be removed.

Configuration/TCA/Overrides/101_sys_template.php
gilbertsoft: You're right, seems not to be useful currently

gilbertsoft: Lets keep it, BP5 is almost here.

Issue #829, opened 2019-12-12
Shortcut in menu always is rendered as inactive


gilbertsoft:
This is a core issue see https://forge.typo3.org/issues/85138

We should provide a fix in the MenuProcessor as long this isn't solved at the TYPO3 core.

Related: #750

PR #830, opened 2019-12-12
[BUGFIX] Circumvent shortcut active state issue in HMENU

gilbertsoft:
Related: #750, https://forge.typo3.org/issues/85138
Resolves: #829

### DIFF
--- a/Classes/DataProcessing/MenuProcessor.php
+++ b/Classes/DataProcessing/MenuProcessor.php
@@ -1,0 +1,102 @@
+<?php
+
+namespace BK2K\BootstrapPackage\DataProcessing;
+
+use TYPO3\CMS\Frontend\ContentObject\ContentObjectRenderer;
+use TYPO3\CMS\Frontend\Page\PageRepository;
+
+/**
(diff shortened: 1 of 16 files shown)


benjaminkott:
@gilbertsoft would not circumvent the issue as it´s a core bug. should be fixed in core and not add the processor again in the bp.

gilbertsoft:
> 
> 
> @gilbertsoft would not circumvent the issue as it´s a core bug. should be fixed in core and not add the processor again in the bp.

Have seen the closed PR? I'll have another look at the core issue, but I don't think Benni will allow a back port to 8.7. So this is finally meant for version not including the core fix. I'm absolutely with you but this seems better than the suggested fix in fluid

gilbertsoft:
Created now two patches for the core to solve this issue:
- https://review.typo3.org/c/Packages/TYPO3.CMS/+/64569
- https://review.typo3.org/c/Packages/TYPO3.CMS/+/64570

gilbertsoft:
Will close this one in favor of the core patche which hopefully goes into the next releases...

Issue #861, opened 2020-03-17
TCA is not compatible to some databases


andreasfernandez:
# Bug Report

Disclaimer: initially reported in Slack, see https://typo3.slack.com/archives/C024TUMM7/p1584426732249000

## Prerequisites

* [ ] Can you reproduce the problem on TYPO3 v8.7 LTS
* [X] Can you reproduce the problem on TYPO3 v9.5 LTS
* [X] Did you [perform a cursory search](https://github.com/issues?q=is%3Aissue+user%3Abenjaminkott+repo%3Abenjaminkott%2Fbootstrap_package)
  to see if your bug or enhancement is already reported?

## Description

If TYPO3 is configured to use PostgreSQL, some queries may fail due to incompatible WHERE clauses in TCA.

Example:
`tx_bootstrappackage_timeline_item/tt_content/foreign_table_where` must be at least `AND tt_content.pid=###CURRENT_PID### AND tt_content.{#CType}="timeline"` to be compatible with databases fiddling with column casing.

### Expected behavior

Creating a tx_bootstrappackage_timeline_item record works fine.

### Actual behavior

Creating such record fails due to database implications.

### Screenshots

<img width="660" alt="Bildschirmfoto 2020-03-17 um 07 30 15" src="https://user-images.githubusercontent.com/1787983/76833547-d2ef4500-682b-11ea-80c0-32ac7584e3cf.png">

PR #864, opened 2020-03-18
[TASK] Provide PostgreSQL DDEV configuration

gilbertsoft:
Tasks:
- [ ] Make volume persistence working

### DIFF
--- a/.ddev/commands/postgres/pgsql_export
+++ b/.ddev/commands/postgres/pgsql_export
@@ -1,0 +1,7 @@
+#!/bin/bash
+
+## Description: Dump the postgresql database to .ddev/import-db/postgresql.db.sql
+## Usage: pgsql_export
+## Example: ddev pgsql_export
+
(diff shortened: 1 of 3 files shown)


gilbertsoft:
@rfay followed https://github.com/drud/ddev-contrib/tree/master/docker-compose-services/postgres here and tried to add persistence which doesn't work currently. Do I have to use a own Docker file to get this working? Any hint? I simply added the data path of PostgreSQL to a volume.

rfay:
For Postgres persistence you'll want to look at https://github.com/drud/ddev-contrib/pull/48 - it adds persistence. Your review there would be appreciated!

gilbertsoft:
Thanks for the hint, that's what I tried but did not work. Will check again and ofc also the PR...

gilbertsoft:
@benjaminkott I'm not sure how to continue with this. Needed it yesterdays to check and fix an issue with PostgreSQL see #861. But I'm not so happy always pulling this container when it's not used. So perhaps we should store this some where else for later use...

benjaminkott:
I currently see no need for this, we drop it for now.

rfay:
Note that DDEV now has native PostgreSQL support, with persistence, everything. It works just like everything else. So this was probably pretty obsolete anyway. https://ddev.readthedocs.io/en/latest/users/extend/database-types/

PR #866, opened 2020-03-24
[BUGFIX] Resolve SCSS files correctly when using symlinked paths

helhum:
When an extension is symlinked, e.g. by using Composer path repositories,
importing scss files from other extensions can fail.

This is now fixed in two ways. The backwards compatible way properly
resolves the path to the final file using the visual path instead
of the real path. for future imports, the extension path is added as import path
so that imports can be done relative to this directory as well.

As a last option for users only relying on the PHP parser,
files can also be imported using the well know TYPO3 syntax
EXT:ext_key/Resources/Public/Scss/my.scss

Closes #585


helhum:
This is meant as alternative to #585 with a similar approach, but does not involve parsing the source, but using library API to resolve paths correctly

gilbertsoft:
Great, thanks! We'll have a look a.s.a.p.

gilbertsoft:
@ohader please have a look at Helmut's approach for the sitepackage in path repo issue.

ohader:
> @ohader please have a look at Helmut's approach for the sitepackage in path repo issue.

Nice, will give it a try later/tonight with my demo projects (e.g. https://github.com/HofUniversityIndieOne2019/book-store-project)

Classes/Parser/ScssParser.php
```diff
@@ -83,9 +85,36 @@ protected function parseFile($file, $settings)
                 'sourceMapBasepath' => '<PATH DOES NOT EXIST BUT SUPRESSES WARNINGS>'
             ]);
         }
+        $absoluteFilename = GeneralUtility::getFileAbsFileName($file);
+        // Adds visual directory path of the initial file as import path
+        // This scenarios happens, when e.g. developing packages using the `path`
+        // repository feature of Composer - having one package in `public/typo3conf/ext/`
+        // and the other one symlinked in e.g. `packages/`.
+        // Since the PHP SCSS parser works on resolved real paths, the symlinked context is lost.
+        $visualImportPath = dirname(GeneralUtility::getFileAbsFileName($file));
```
helhum, line 94: `$absoluteFilename` can be used here

gilbertsoft: Yeah, saw this too. Thinking also about to rename $visualImportPath because I'm using this variable too for my patch #868 and in this context the naming is wrong. But I thought about streamlining this after the merge of this two patches...

benjaminkott:
Hey everyone, thx a lot for your work in this area - i tried to build a test case so we can validate it´s working correctly and to be prepared for changes in scssphp. Could you have a look and extend/correct my test cases?

Classes/Parser/ScssParser.php
```diff
@@ -83,9 +85,36 @@ protected function parseFile($file, $settings)
...
+        $scss->addImportPath(function ($url) use ($visualImportPath) {
+            // Resolve potential back paths manually using PathUtility::getCanonicalPath,
+            // but make sure we do not break out of TYPO3 application path using GeneralUtility::getFileAbsFileName
+            // Also resolve EXT: paths if given
+            $isTypo3Absolute = strpos($url, '..') === false;
```
gilbertsoft, line 99: @helhum I had to change this line because there are also cases with absolute paths and they failed. Not sure if we should use `../` here for relative path detection or the dots are enough.

helhum: what cases with absolute paths failed?

Classes/Parser/ScssParser.php
```diff
@@ -83,9 +85,36 @@ protected function parseFile($file, $settings)
...
+        // This scenarios happens, when e.g. developing packages using the `path`
+        // repository feature of Composer - having one package in `public/typo3conf/ext/`
+        // and the other one symlinked in e.g. `packages/`.
+        // Since the PHP SCSS parser works on resolved real paths, the symlinked context is lost.
+        $visualImportPath = dirname($absoluteFilename);
+        $scss->addImportPath(function ($url) use ($visualImportPath) {
+            // Resolve potential back paths manually using PathUtility::getCanonicalPath,
+            // but make sure we do not break out of TYPO3 application path using GeneralUtility::getFileAbsFileName
+            // Also resolve EXT: paths if given
+            $isTypo3Absolute = !(strpos($url, '..') === 0);
```
helhum, line 99: that is definitely a wrong check here now

gilbertsoft: changed as discussed

helhum:
looks good now

PR #868, opened 2020-04-04
[BUGFIX] Fix paths in url() statements

### LINKED ISSUE #607: Wrong path with sitepackage for Photoswipe Lightbox Skin images
# Bug Report

## Prerequisites

* [ ] Can you reproduce the problem on TYPO3 v8.7 LTS
* [x] Can you reproduce the problem on TYPO3 v9.5 LTS
* [x] Did you [perform a cursory search](https://github.com/issues?q=is%3Aissue+user%3Abenjaminkott+repo%3Abenjaminkott%2Fbootstrap_package) to see if your bug or enhancement is already reported?

## Description
With a sitepackage generated at https://www.sitepackagebuilder.com/ the path for Images of the lightbox is wrong and there is a 404 error in Console
https://www.example.com/typo3conf/ext/sitepackage/Resources/Public/Images/PhotoSwipe/default-skin.png
## Steps to Reproduce

1. Install sitepackage
2. Add several Images with "Enlarge on Click"
3. open the content in Frontend on a Desktop PC
4. open the lightbox
5. try to go forward to the next image

**Expected behavior:** 
some icons to browse to the next image

**Actual behavior:** 
nothing

## Versions

TYPO3 v.9.5.1, PHP 7.2.10 MySQL 5.7.22

gilbertsoft:
This patch fixes wrong paths in .scss files url() statements if the bootstrap_package theme is included by a sitepackage. It first tries to resolves paths to the sitepackage, second to the bootstrap_package and if already not found keeps the current value.

Fixes #607


Classes/Parser/ScssParser.php
```diff
@@ -83,10 +84,36 @@ protected function parseFile($file, $settings)
...
         }
-        $css = $scss->compile('@import "' . $file . '"');
 
         $absoluteFilename = GeneralUtility::getFileAbsFileName($file);
-        $relativePath = $settings['cache']['tempDirectoryRelativeToRoot'] . dirname(substr($absoluteFilename, strlen($this->getPathSite()))) . '/';
+
+        // Make paths in url() statements relative to site root
+        $absoluteFilePath = dirname($absoluteFilename);
+        $siteFilePath = dirname(substr($absoluteFilename, strlen($this->getPathSite())));
+        $absoluteBootstrapPackagePath = GeneralUtility::getFileAbsFileName(ExtensionManagementUtility::extPath('bootstrap_package') . 'Resources/Public/Scss/Theme');
```
helhum, line 93: `ExtensionManagementUtility::extPath` returns the absolute path already. No need to pass it to `GeneralUtility::getFileAbsFileName` additionally.
I'd name the variable to something that represents what is meant, e.g. `$bootstrapPackageThemePath`

Classes/Parser/ScssParser.php
```diff
@@ -83,10 +84,36 @@ protected function parseFile($file, $settings)
...
             ]);
         }
-        $css = $scss->compile('@import "' . $file . '"');
 
         $absoluteFilename = GeneralUtility::getFileAbsFileName($file);
-        $relativePath = $settings['cache']['tempDirectoryRelativeToRoot'] . dirname(substr($absoluteFilename, strlen($this->getPathSite()))) . '/';
+
+        // Make paths in url() statements relative to site root
+        $absoluteFilePath = dirname($absoluteFilename);
+        $siteFilePath = dirname(substr($absoluteFilename, strlen($this->getPathSite())));
```
helhum, line 92: There is API to get the web path of an absolute path: `\TYPO3\CMS\Core\Utility\PathUtility::getAbsoluteWebPath`

helhum: I personally also think that deriving some public asset dir from a (should be private source dir) isn't optimal. At least I'd name the variable to something that represents what is meant, e.g. `$sitePackageThemeWebPath`

Classes/Parser/ScssParser.php
```diff
@@ -83,10 +84,36 @@ protected function parseFile($file, $settings)
...
+        $absoluteBootstrapPackagePath = GeneralUtility::getFileAbsFileName(ExtensionManagementUtility::extPath('bootstrap_package') . 'Resources/Public/Scss/Theme');
+        $siteBootstrapPackagePath = substr($absoluteBootstrapPackagePath, strlen($this->getPathSite()));
```
helhum, line 94: I'd name the variable to something that represents what is meant, e.g. `$bootstrapPackageThemeWebPath` and use the API method (see above) to derive it from absolute path

Classes/Parser/ScssParser.php
gilbertsoft: I don't know why Benji used here the Public path for scss and less, imho this could go into Private and the final css should be public. But that's an other patch...

Classes/Parser/ScssParser.php
```diff
@@ -114,12 +114,44 @@ protected function parseFile($file, $settings)
...
         $css = $scss->compile('@import "' . $absoluteFilename . '"');
 
-        // Fix url() statements
-        $relativePath = $settings['cache']['tempDirectoryRelativeToRoot'] . dirname($settings['file']['relative']) . '/';
+        // Fix paths in url() statements to be relative to temp directory
+        $relativeTempPath = $settings['cache']['tempDirectoryRelativeToRoot'];
         $search = '%url\s*\(\s*[\\\'"]?(?!(((?:https?:)?\/\/)|(?:data:?:)))([^\\\'")]+)[\\\'"]?\s*\)%';
-        $replace = 'url("' . $relativePath . '$3")';
+        $replace = 'url("' . $relativeTempPath . '$3")';
         $css = preg_replace($search, $replace, $css);
```
gilbertsoft, line 155: I think I should add this to the function above too...

gilbertsoft: Not possible, has to be done after compilation

PR #876, opened 2020-04-30
[TASK] Add typo3fluid/fluid to composer dependencies

tomasnorre:
# Pull Request

## Prerequisites

## Description

For me it looks like the dependency for `typo3fluid/fluid` is missing in composer e.g. the InlineSvgViewHlper is requiring the `TYPO3Fluid\Fluid\Core\ViewHelper\AbstractViewHelper`.
https://github.com/benjaminkott/bootstrap_package/blob/master/Classes/ViewHelpers/InlineSvgViewHelper.php#L17

It can be that most people have the `typo3fluid/fluid` include in their projects, but as it's used here directly it should in my opinion be a hard dependency.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -46,7 +46,7 @@
         "typo3/cms-rte-ckeditor": "^9.5 || ^10.0 || 10.*@dev",
         "typo3/cms-seo": "^9.5 || ^10.0 || 10.*@dev",
         "scssphp/scssphp": "^1.0.6",
-        "typo3fluid/fluid": "^2.6"
+        "typo3fluid/fluid": "^2.5.2"
     },
     "require-dev": {
         "typo3/cms-felogin": "^9.5 || ^10.0 || 10.*@dev",
--- a/composer.json
+++ b/composer.json
@@ -45,7 +45,8 @@
         "typo3/cms-install": "^9.5 || ^10.0 || 10.*@dev",
         "typo3/cms-rte-ckeditor": "^9.5 || ^10.0 || 10.*@dev",
         "typo3/cms-seo": "^9.5 || ^10.0 || 10.*@dev",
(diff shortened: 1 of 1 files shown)


gilbertsoft:
Thanks for this. Yes and there are other dependencies missing too. Had it already on my task list. One less :)

gilbertsoft:
First one I never heard before, will check. The second one is in my composer global space.

composer.json
```diff
@@ -45,7 +45,8 @@
         "typo3/cms-install": "^9.5 || ^10.0 || 10.*@dev",
         "typo3/cms-rte-ckeditor": "^9.5 || ^10.0 || 10.*@dev",
         "typo3/cms-seo": "^9.5 || ^10.0 || 10.*@dev",
-        "scssphp/scssphp": "^1.0.6"
+        "scssphp/scssphp": "^1.0.6",
+        "typo3fluid/fluid": "^2.6"
```
gilbertsoft, line 49: Can you please change the dependency to `^2.5.2` which is used for 9.5.0.

georgringer: I would avoid such change and keep the dependency through typo3. eg. this  change prohibits the usage of fluid 3.0

tomasnorre: As there is a hard dependency for this in this extension, it should be listed here as well so I would suggest to do a 
```
"typo3fluid": "^2.5.2 || ^3.0" 
``

or something depending on which fluid version works with this.

gilbertsoft: Thx for the hint @georgringer, so your new suggestion is the way to go @tomasnorre. Omitting a dependency is imho not smart...

gilbertsoft: Benji just told me, BP is not fully compatible with Fluid 3.x. So I'll merge it now and we'll fix it for Fluid 3.x later. Thanks you both!

PR #886, opened 2020-05-18
[DOCS] Add quick start chapter

gilbertsoft:


gilbertsoft:
Rendered version is available at https://quick-start.web02.opsone.core.gilbertsoft.net/QuickStart/Index.html

Documentation/QuickStart/Preparation/Composer.rst
```diff
@@ -0,0 +1,70 @@
...
+
+TYPO3 Core Extensions
+=====================
+
+Now it's time to tell Composer which packages and TYPO3 extensions you like to
+install and setup. For a initial setup run the following commands:
+
+.. code-block:: bash
+
+   ddev composer create "typo3/cms-base-distribution:^10" --no-install
```
benjaminkott, line 22: Would avoid version numbers in documentation.

sypets: I like your approach very much, I am just not sure here is the right place for it. 

This goes way beyond the bootstrap_package. Do you think it might be a good idea to add this to the official [Installation Guide](https://docs.typo3.org/m/typo3/guide-installation/master/en-us/) instead? I think that is where people would look for help on installing TYPO3 and I am not a huge fan of fragmented information in various places.

Disclaimer - I sufficiently looked at the content - I may have misjudged this.

I think this would need some structural decisions, but I have also been thinking for a while, if the building block approach in the installation guide is such a good idea or if it might be better to have one page full installation instructions for several typical usecases, e.g. with ddev / composer / bootstrap package, without composer etc. However I find it difficult to come up with a solution that does not have several pros and cons.

To illustrate this:

1. building block approach: one page for each individual step, sometimes several options available, the reader must find his path through the steps (used currently)
2. One page with one clear objective, e.g. install TYPO3 with DDEV, composer and introduction package. That would be more like your quickstart here. 

Currently we have individual pages for various steps, e.g. install package, Admin tool. You are then walked through the various steps and often have something like, if you have Composer, do this next, otherwise do this etc.

What I am also missing in the official Installation guide are some steps which are strictly needed, such as site configuration.

PR #888, opened 2020-05-19
[BUGFIX] Make gallery compatible with TYPO3 10.4.x

gilbertsoft:


dahaupt:
@gilbertsoft Using that approach the gallery itself is rendered but the pagination does not work because the extbase context is missing. I've also created pull request (#890), tested in v10 only. 

I've linked the corresponding TYPO3 issue within an comment as I had a hard time debugging the issue with the pagination. You can ping me in Slack if you have any questions!

gilbertsoft:
I'm not sure why it seemed to work here. But as mentioned in your PR I will use the native paginator now available since TYPO3 10 if possible.

benjaminkott:
We are taking https://github.com/benjaminkott/bootstrap_package/pull/890 instead of this one. thx for the PR.

PR #948, opened 2020-11-12
[BUGFIX] Apply latest scssphp changes to import behavior

### LINKED ISSUE #947: Path issues with scssphp 1.4
After updating scssphp package to 1.4, scss paths doesn't work correctly. Custom scss path from sitepackage typescript causes fatal errors.
Possible because of symlinks, haven't tested without symlink to public folder.

Same issue with 11.0.3 dev branch.

Not sure if this is a bug with scssphp compiler or bootstrap_package path detection.

Downgrading to scssphp 1.3 resolves the issues.

Apologies for the lack of details, currently very busy so figured it was more important to get the bug reported.

gilbertsoft:
# Pull Request

## Related Issues

* Fixes #947

## Prerequisites

* [x] Changes have been tested on TYPO3 v8.7 LTS
* [x] Changes have been tested on TYPO3 dev-master
* [x] Changes have been tested on PHP 7.0.x
* [x] Changes have been tested on PHP 7.1.x

## Description

This patch fixes our own scss import behavior with the latest changes at scssphp/scssphp version 1.4. Also the expectation of the CSS output is fixed.

## Steps to Validate

Run the functional tests with scssphp 1.4 and a lower release e.g. 1.3

The unit tests will fail with versions lower than 1.4 because of an other fix in scssphp 1.4.

### DIFF
--- a/Classes/Parser/ScssParser.php
+++ b/Classes/Parser/ScssParser.php
@@ -96,7 +96,7 @@ class ScssParser extends AbstractParser
             // Resolve potential back paths manually using PathUtility::getCanonicalPath,
             // but make sure we do not break out of TYPO3 application path using GeneralUtility::getFileAbsFileName
             // Also resolve EXT: paths if given
-            $isTypo3Absolute = (strpos($url, 'EXT:') === 0) || (strpos($url, '/') === 0);
+            $isTypo3Absolute = (strpos($url, 'EXT:') === 0) || PathUtility::isAbsolutePath($url);
             $fileName = $isTypo3Absolute ? $url : $visualImportPath . '/' . $url;
             $full = GeneralUtility::getFileAbsFileName(PathUtility::getCanonicalPath($fileName));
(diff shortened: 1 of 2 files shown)


Classes/Parser/ScssParser.php
```diff
@@ -96,7 +96,7 @@ protected function parseFile($file, $settings)
             // Resolve potential back paths manually using PathUtility::getCanonicalPath,
             // but make sure we do not break out of TYPO3 application path using GeneralUtility::getFileAbsFileName
             // Also resolve EXT: paths if given
-            $isTypo3Absolute = strpos($url, 'EXT:') === 0;
+            $isTypo3Absolute = (strpos($url, 'EXT:') === 0) || (strpos($url, '/') === 0);
```
stof, line 99: be careful, absolute paths don't start with `/` on Windows.

gilbertsoft: Yeah, you're right. Thanks for the hint!

PR #911, opened 2020-07-24
[!!!][TASK] Remove TYPO3 9.5 and add 11 support 

gilbertsoft:
Preparation for TYPO3 v11 support.

Pending task (@benjaminkott):

- [x] Deployment in Travis is still missing in GH Actions
- [x] Travis has to be removed or adapted properly
- [x] Scrutinizer has to be removed or adapted properly
- [x] Expected checks for 9 have to be removed and added for 11 in GH settings

PR #971, opened 2021-02-14
Make bootstrap package availible for first release of TYPO3 11

linawolf:
The requirements still only allow "11.*.*@dev" even though there is the release


jonaseberle (approved):

gilbertsoft:
Thanks for your PR. I'm preparing this since months see https://github.com/benjaminkott/bootstrap_package/pull/911.

jonaseberle (changes requested):
`^11.0 || ^11.1` is superfluous since one is a subset of the other.
The `11.*.*@dev`  should be kept to allow using it with non-released versions.

linawolf:
It is availible now, So this PR can be closed

PR #1002, opened 2021-05-10
[BUGFIX] Change extension category to distribution

spoonerWeb:
As there is a category for distributions, it would also appear in the distribution extension list in extension manager.


spoonerWeb:
Okay, thought it's also a Distribution. If not, I need to read the definition again. 😄

benjaminkott:
I agree with @gilbertsoft the bp is not a distribution as it delivers no site config, I will close this pr - thx for the suggestion.

PR #1004, opened 2021-05-21
[TASK] PHPStan clean ups

benjaminkott:


gilbertsoft:
This patch reverted changes of #984 for the TYPO3 11 compatibility, will reapply it...

PR #1010, opened 2021-06-09
[TASK] Disable deployment and publish workflows for forks

gilbertsoft:
Merging changes to forks leads to workflow errors and these new conditions avoid it.

grafik


gilbertsoft:
Much better now :)

grafik

PR #1023, opened 2021-07-23
[BUGFIX] SassError: compound selectors may no longer be extended.

davidzach:
SassError: compound selectors may no longer be extended. Consider `@extend .form-control, .is-invalid` instead. See: https://sass-lang.com/documentation/breaking-changes/extend-compound for details.


gilbertsoft:
Thanks for your fix!

benjaminkott:
Hi all, i have to remove this patch again since it´s incomplete and the rendered differs, I know about this deprecation but extending `.form-control, .is-invalid` is != `.form-control.is-invalid`. SCSS patches also must include the compiled CSS, the changes would have been visible there.

PR #1037, opened 2021-08-11
[BUGFIX] Update doctrine usages

benjaminkott:


gilbertsoft:
@benjaminkott The doctrine errors can be fixed by using phpstan/phpstan-doctrine btw. I can step in if you like and fix it...

benjaminkott:
> @benjaminkott The doctrine errors can be fixed by using phpstan/phpstan-doctrine btw. I can step in if you like and fix it...

cool will check

Issue #1076, opened 2021-09-01
Wrong image size in nested contentcontainer


buepro:
# Bug Report

## Prerequisites

* [x] Can you reproduce the problem on TYPO3 v10.4 LTS
* [ ] Can you reproduce the problem on TYPO3 v11.3
* [x] Did you [perform a cursory search](https://github.com/issues?q=is%3Aissue+user%3Abenjaminkott+repo%3Abenjaminkott%2Fbootstrap_package)
  to see if your bug or enhancement is already reported?

## Description

When contentcontainer elements are nested the image calculation isn't correct.

## Steps to Reproduce

1. Add an image to a nested column container where the columns aren't even (e.g. first container with 25%-75% and second container with 75%-25%. Place the image in left column from second container)

## Versions

- master#d513638b

buepro:
Can't reproduce the error with master#3c6c1040.

PR #1077, opened 2021-09-01
[BUGFIX] Correct variants in nested contentcontainer

buepro:
Resolves: #1076


buepro:
Became obsolete since the error can't be reproduced with master#3c6c1040.

PR #1080, opened 2021-09-15
[TASK] Allow pagination template to be overwritten

benjaminkott:


Hawkeye1909:
Hello @benjaminkott,

may I ask why this isn't in v12 yet?

PR #1095, opened 2021-10-01
[TASK] Migrate form definition

gilbertsoft:
See https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/10.0/Feature-80420-AllowMultipleRecipientsInEmailFinisher.html

### DIFF
--- a/Resources/Private/Forms/Contact.form.yaml
+++ b/Resources/Private/Forms/Contact.form.yaml
@@ -8,13 +8,10 @@ finishers:
         identifier: EmailToReceiver
         options:
             subject: '{subject}'
-            recipientAddress: ''
-            recipientName: ''
             senderAddress: '{email}'
             senderName: '{fullname}'
-            replyToAddress: '{email}'
-            carbonCopyAddress: ''
-            blindCarbonCopyAddress: ''
+            replyToRecipients:
+              '{email}': ''
             format: html
             attachUploads: true
     -


Resources/Private/Forms/Contact.form.yaml
```diff
@@ -8,13 +8,10 @@ finishers:
         identifier: EmailToReceiver
         options:
             subject: '{subject}'
-            recipientAddress: ''
-            recipientName: ''
```
mbrodala, line 12: These have been dropped without replacement?

Resources/Private/Forms/Contact.form.yaml
```diff
@@ -8,13 +8,10 @@ finishers:
...
             senderAddress: '{email}'
             senderName: '{fullname}'
```
mbrodala, line 12: These must also be migrated.

Resources/Private/Forms/Contact.form.yaml
```diff
@@ -8,13 +8,10 @@ finishers:
...
-            replyToAddress: '{email}'
-            carbonCopyAddress: ''
-            blindCarbonCopyAddress: ''
```
mbrodala, line 17: These have been dropped without replacement?

mbrodala (changes requested):

Resources/Private/Forms/Contact.form.yaml
gilbertsoft: Just did the same mentioned in Slack. What should this look like?

```
recipients:
  '': ''
```

Resources/Private/Forms/Contact.form.yaml
gilbertsoft: According the sources this looks still fine https://github.com/TYPO3/typo3/blob/10.4/typo3/sysext/form/Classes/Domain/Finishers/EmailFinisher.php#L54-L55

mbrodala: Yes, got confused myself. This of course is still a single value.

Resources/Private/Forms/Contact.form.yaml
mbrodala: As mentioned in Slack these options should be skipped in the first place and thus also do not need a migration. Users of the Bootstrap Package will never edit this file to add static values. Instead they will either copy it and add static values or set recipients via finisher overrides.

Resources/Private/Forms/Contact.form.yaml
mbrodala: So these must be left unchanged.

Resources/Private/Forms/Contact.form.yaml
mbrodala: CC and BCC are optional and should be skipped if no useful static value can be set.

mbrodala (approved):

Resources/Private/Forms/Contact.form.yaml
gilbertsoft: Yes, agree.

PR #1157, opened 2022-04-11
Exclude Tests folder for export

simonschaufi:
* Documentation should be read online and the raw rest files are not extremely useful
* Tests should never be released


gilbertsoft (approved):

gilbertsoft:
Thanks!

PR #1213, opened 2022-10-13
[TASK] Replace deprecated set-output command

gilbertsoft:
See https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/


andreasfernandez (approved):
Looks fine, I currently do the same in all projects :D

PR #1240, opened 2023-01-13
[TASK] add additionalAttributes to FrameViewHelper

baschte:
# Pull Request
Adds the option to add additional attributes to rendered frame div of content elements


m-strotmeier:
I appreciate the idea of making the FrameViewHelper more flexible. 
How about to also add the option to add some additional classes manually? My example would be to add a class called "rounded" to the frame. Workaround is to rename this class to "frame-option-rounded" and use the frame options for this.

benjaminkott:
patch is on master branch on remote, dropping in favor of https://github.com/benjaminkott/bootstrap_package/pull/1268

PR #1277, opened 2023-05-21
Add &display=swap to Google Fonts link

ste101:


benjaminkott:
It´s already present see the diff.

image

PR #1351, opened 2023-06-06
[TASK] Add meaningful label to each second field in a group in extend…

### LINKED ISSUE #1312: A11y Code Sprint Berlin: Success Criterion 9.3.3.2 - add title to second field in form groups
# Feature Request

## Description

If form fields are grouped as they are in the [advanced search](https://bitv-v12.live.typo3.cps.321.works/content-examples/form-elements/search?tx_indexedsearch_pi2%5Baction%5D=form&tx_indexedsearch_pi2%5Bcontroller%5D=Search&tx_indexedsearch_pi2%5Bsearch%5D%5BextendedSearch%5D=1&cHash=bd4ce4122dff0be3360b7c2c0634c9fb), screen readers announce the lable only for the first field. The purpose of the second field can be unclear. 

## Solution
Add meaningful label to each second field in a group. 

image
Labels: "Match" and "Match in"

image
Add label: "Language" 

image
Add label: "Show in order"

schmigotzki:
…ed search form

# Pull Request

## Related Issues

* Resolves #1312

## Prerequisites

* [x] Changes have been tested on TYPO3 v12.4 LTS
* [x] Changes have been tested on PHP 8.1

## Description

Problem:
If form fields are grouped, every field needs his own label. Otherwise the purpose of the second field can be unclear.

Solution:
Add meaningful label to each second field in a group.

## Steps to Validate

1. Open the advanced search form in the frontend
2. Check markup of form groups in the advanced search form

### DIFF
--- a/Resources/Private/Language/locallang.xlf
+++ b/Resources/Private/Language/locallang.xlf
@@ -105,6 +105,16 @@
                 <source>Summary</source>
             </trans-unit>
 
+            <trans-unit id="searchForm.element.defaultOperand.label">
+                <source>Match in</source>
+            </trans-unit>
+            <trans-unit id="searchForm.element.language.label">
+                <source>Language</source>
+            </trans-unit>
+            <trans-unit id="searchForm.element.sortDesc.label">
+                <source>Show in order</source>
+            </trans-unit>
+
             <trans-unit id="cookieconsent.common.header">
                 <source>Cookies used on the website!</source>
             </trans-unit>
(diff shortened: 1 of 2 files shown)


benjaminkott (changes requested):
Aside from template changes, the labels should be changed in the core and not added to the bootstrap_package. Since we cannot have a release before the A11Y check is done, we can add them temporarily to the package and remove them later.

PR #1442, opened 2023-11-09
Fix render textpic above

### LINKED ISSUE #1441: Render gallery container in TextpicAbove and TextmediaAbove only if files exist
# Feature Request

## Is your feature request related to a problem? Please describe

Sometimes editors add CE like textpic but do not add an image at all. Since the default setting for 'imageorient' is 'Above, center' (0) and due to the fact a gap is added between gallery and text ([here](https://github.com/benjaminkott/bootstrap_package/blob/6f3e3c2b95fa82171be970268626dbda79bbea81/Resources/Public/Scss/components/_textpic.scss#L8)), there is a not needed space above the text when the container is rendered.

## Describe the solution you'd like

Only render the gallery container in the files TextpicAbove.html and TextmediaAbove.html if files exist to prevent the gap.

Hawkeye1909:
# Pull Request

## Related Issues

* Resolves #1441

## Prerequisites

* [x] Changes have been tested on TYPO3 v11.5 LTS
* [ ] Changes have been tested on TYPO3 v12.4 LTS

## Description

Check if there are files before rendering the gallery container to prevent unused space above text.

## Steps to Validate

1. Add Textpic CE without addding an image

### DIFF
--- a/Resources/Private/Templates/ContentElements/TextmediaAbove.html
+++ b/Resources/Private/Templates/ContentElements/TextmediaAbove.html
@@ -4,9 +4,11 @@
 <f:section name="Main">
 
     <div class="textmedia textmedia-above">
-        <div class="textmedia-item textmedia-gallery">
-            <f:render partial="Media/Gallery" arguments="{_all}" />
-        </div>
+        <f:if condition="{files}">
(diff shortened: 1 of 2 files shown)


benjaminkott:
thx for the patch!

PR #1460, opened 2024-01-07
Simplify language and footer meta menu: Use list-inline instead of custom styles

SventB:
# Pull Request

## Prerequisites

* [ ] Changes have been tested on TYPO3 v11.5 LTS
* [x] Changes have been tested on TYPO3 v12.4 LTS
* [ ] Changes have been tested on PHP 7.4
* [ ] Changes have been tested on PHP 8.0
* [ ] Changes have been tested on PHP 8.1
* [x] Changes have been tested on PHP 8.2
* [ ] CSS has been rebuilt (only if there are SCSS changes `cd Build; npm ci; npm run build`)

## Description

Simplifies language and footer meta menu by using Bootstrap class "list-inline" (nothing else is the language and footer meta menu right now) instead of custom styles.

## Steps to Validate

Check language and footer meta menu.

### DIFF
--- a/Resources/Private/Partials/Page/Navigation/Language.html
+++ b/Resources/Private/Partials/Page/Navigation/Language.html
@@ -1,16 +1,16 @@
 <html xmlns:f="http://typo3.org/ns/TYPO3/CMS/Fluid/ViewHelpers" data-namespace-typo3-fluid="true">
 <f:if condition="{languagenavigation}">
-    <ul id="language_menu" class="language-menu">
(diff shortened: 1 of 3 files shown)


benjaminkott:
@SventB i understand where this demand is comming from. anyways i would like to keep certain elements to not use bootstrap at all to keep compat moving forward to avoid breaking on bootstrap updates. I still think this makes customizing more easy of we keep the styling seperate.

PR #1501, opened 2024-07-28
Fix image width calculation in ImageVariantsUtility.php

### LINKED ISSUE #1500: Wrong image width calculation in ImageVariantsUtility.php
# Bug Report

## Prerequisites

* [X] Can you reproduce the problem on TYPO3 v11.5 LTS
* [X] Can you reproduce the problem on TYPO3 v12.4 LTS
* [X] Did you [perform a cursory search](https://github.com/issues?q=is%3Aissue+user%3Abenjaminkott+repo%3Abenjaminkott%2Fbootstrap_package)
  to see if your bug or enhancement is already reported?

## Description

Once again, a brilliant mathematician has pulled off a stroke of genius and caused the widths of the images not to be calculated correctly:

For example, the CSS states:
```css
.textmedia-item, .textpic-item {
	width: calc(50% -(40px / 2));
}
```

Which, for example, leads to an image size of 620px with a container width of 1280px. logical, because $1280 * 0.5 - 40 / 2 = 620$

PHP produces the same result in this case, as the gutters are removed first and then the multiplier is applied.
Mathematically a different way, but in this case the same result: $(1280 - 40) * 0.5 = 620$

However, if a different multiplier is used, such as 25% instead of 50%, the situation is unfortunately different.
CSS calculates: $1280 * 0.25 - 40 / 2 = 300$
And PHP calculates: $(1280 - 40) * 0.25 = 310$

**Ergo:**
If you want to reflect the calculation of image/text widths used in SCSS in PHP code, you would first have to apply the multiplier and then subtract half the gutter.

stephankellermayr:
Fixes #1500

### DIFF
--- a/Classes/Utility/ImageVariantsUtility.php
+++ b/Classes/Utility/ImageVariantsUtility.php
@@ -67,11 +67,11 @@ class ImageVariantsUtility
         $variants = $variants !== null ? $variants : [];
         $variants = self::processVariants($variants);
         $variants = self::processResolutions($variants);
-        if ($gutters !== null) {
-            $variants = self::removeGutters($variants, $gutters);
-        }
         if ($multiplier !== null) {
             $variants = self::processMultiplier($variants, $multiplier);
+        }
+        if ($gutters !== null) {
+            $variants = self::removeGutters($variants, $gutters);
         }
         if ($corrections !== null) {
             $variants = self::processCorrections($variants, $corrections);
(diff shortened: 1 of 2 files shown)


stephankellermayr:
Great, since the bugs are piling up here and I was careless, PR #1453 has also crept in here.

Anyway, probably nobody will take care of it in the foreseeable future anyway...

PR #1582, opened 2025-07-22
[FEATURE] Introduce event to modify icon providers

buchmarv:
# Pull Request

## Prerequisites

* [x] Changes have been tested on TYPO3 v13.4 LTS
* [ ] Changes have been tested on TYPO3 v14
* [ ] Changes have been tested on PHP 8.2
* [ ] Changes have been tested on PHP 8.3
* [x] Changes have been tested on PHP 8.4
* [-] CSS has been rebuilt (only if there are SCSS changes `cd Build; npm ci; npm run build`)

## Description

This pull requests introduces a PSR-14 event `ModifyIconProvidersEvent` which can be used to modify the registered icon providers. This also allows using anonymous classes as icon providers and DI in icon provider classes.
Also it tags classes implementing `IconProviderInterface`, registering them as icon providers. This removes the need to register them manually in `ext_localconf.php`.
The existing icon providers are migrated to use the new event.

## Steps to Validate

a) Create a class implementing `IconProviderInterface`. It will be registered automatically via a DI tag.

b) [Create an event listener](https://docs.typo3.org/permalink/t3coreapi:eventdispatcherimplementation) inside an extension, listening for the `ModifyIconProvidersEvent` event. Modify the event's icon providers and validate the result.

### DIFF
--- a/Classes/Service/IconProviderService.php
+++ b/Classes/Service/IconProviderService.php
@@ -25,7 +25,7 @@ final readonly class IconProviderService
     public function __invoke(ModifyIconProvidersEvent $event): void
     {
         $iconProviders = $event->getIconProviders();
-        /** @var IconProviderInterface[] $iconProvider */
+        /** @var IconProviderInterface $iconProvider */
         foreach ($this->iconProviders as $iconProvider) {
             $iconProviders[] = $iconProvider;
         }
--- a/Classes/Service/IconProviderService.php
+++ b/Classes/Service/IconProviderService.php
@@ -11,24 +11,23 @@ declare(strict_types = 1);
 namespace BK2K\BootstrapPackage\Service;
 
 use BK2K\BootstrapPackage\Events\ModifyIconProvidersEvent;
-use BK2K\BootstrapPackage\Icons\GlyphiconsProvider;
 use BK2K\BootstrapPackage\Icons\IconProviderInterface;
-use BK2K\BootstrapPackage\Icons\IoniconsProvider;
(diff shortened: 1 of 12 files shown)


Configuration/Services.yaml
```diff
@@ -13,6 +13,12 @@ services:
         identifier: 'bk2k/bootstrap-package/set-backend-styling'
         event: TYPO3\CMS\Core\Package\Event\AfterPackageActivationEvent
 
+  BK2K\BootstrapPackage\Service\IconProviderService:
+    tags:
+      - name: event.listener
+        identifier: 'bk2k/bootstrap-package/add-icon-providers'
```
benjaminkott, line 19: this should be modify right?

Classes/Service/IconProviderService.php
```diff
@@ -0,0 +1,36 @@
...
+
+use BK2K\BootstrapPackage\Events\ModifyIconProvidersEvent;
+use BK2K\BootstrapPackage\Icons\GlyphiconsProvider;
+use BK2K\BootstrapPackage\Icons\IconProviderInterface;
+use BK2K\BootstrapPackage\Icons\IoniconsProvider;
+use TYPO3\CMS\Core\Utility\GeneralUtility;
+
+final class IconProviderService
+{
+    private array $iconProvidersToAdd = [
```
benjaminkott, line 21: we should tag the providers as IconProvider instead of hardcoding them here - as initially discussed

benjaminkott (changes requested):

PR #1597, opened 2025-11-17
[TASK] Make ViewHelper "render" declaration compatible to fluid5

garvinhicking:
This is a POC. It works for me, I have not yet went through other tests/routines. I *think* it *should* work with older Fluid versions.


benjaminkott:
resolved in fluid, not changes needed

PR #1599, opened 2025-12-16
[TASK] Add referrerpolicy attribute to media

Patta:
## Description

Adds `referrerpolicy` to embedded media. A site that sends
`Referrer-Policy: same-origin` strips the referrer from the embed request,
and YouTube answers that with error 153. The element level attribute
overrides the document policy for that one fetch and brings the referrer
back.

The value comes from a new site set setting,
`plugin.bootstrap_package_contentelements.media.additionalAttributes.referrerpolicy`,
defaulting to `strict-origin-when-cross-origin` — the browser default, so a
site that sends no header of its own renders as before. An empty value
renders no attribute and leaves the referrer policy of the site in place.

`referrerpolicy` is defined on `iframe` and not on `video`. The media
rendering partial serves online media and locally hosted video alike, so it
asks `OnlineMediaHelperRegistry` which of the two it has and hands the
attribute over only for the former. The external media element builds its
iframe itself and reads the same setting.

## Type of change

* [x] Bugfix
* [x] Task
* [ ] Feature
* [ ] Documentation

## Related issues

None.

## How to validate

1. Send `Referrer-Policy: same-origin` from the web server.
2. Reference a YouTube video in a media element: the iframe carries
   `referrerpolicy="strict-origin-when-cross-origin"` and the player loads
   instead of failing with error 153.
3. Reference an mp4 in a textmedia element: the `video` tag carries no
   `referrerpolicy`.
4. Empty the setting: neither carries the attribute.
5. `composer test` — `Tests/Functional/ContentElements/MediaTest.php` covers
   1 to 4 and fails without the change to the rendering partial.

## AI assistance

* Agent and version: Claude Code
* Model and effort/reasoning level: Claude Opus 5, high
* Share written by the agent: the first commit is hand written by the
  contributor. The three follow-up commits — the restriction to online
  media, the external media element and the setting description, with their
  tests — were written by the agent from a review of the first.
* Reviewed and understood before pushing: yes, by the maintainer

## Checklist

* [x] Targets `master`
* [x] Commits are signed off (`git commit -s`)
* [x] Commit subjects follow `[BUGFIX|TASK|FEATURE] Subject`
* [x] `composer cgl` leaves the code unchanged
* [x] `composer phpstan` passes
* [x] `composer test` passes
* [x] A bugfix comes with a test that fails without it — appreciated,
      not required
* [x] Holds on every TYPO3 and PHP version declared in `composer.json`
* [ ] Assets rebuilt and committed if SCSS, JavaScript or icons changed
      (`npm --prefix Build ci && npm --prefix Build run build`)
* [x] Documentation updated if the change is user facing

### DIFF
--- a/Classes/Utility/ExternalMediaUtility.php
+++ b/Classes/Utility/ExternalMediaUtility.php
@@ -29,7 +29,7 @@ class ExternalMediaUtility
      * Get the embed code for the given url if possible
      * and add a css class on the iframe
      */
-    public function getEmbedCode(?string $url, ?string $class, ?string $title = null): ?string
+    public function getEmbedCode(?string $url, ?string $class, ?string $title = null, ?string $referrerPolicy = null): ?string
     {
         if ($url === null || $url === '') {
             return null;
@@ -51,6 +51,9 @@ class ExternalMediaUtility
                 }
                 if ($class !== null && trim($class) !== '') {
                     $attributes['class'] = trim($class);
+                }
+                if ($referrerPolicy !== null && trim($referrerPolicy) !== '') {
+                    $attributes['referrerpolicy'] = trim($referrerPolicy);
                 }
                 return '<iframe ' . GeneralUtility::implodeAttributes($attributes, true) . ' allowfullscreen></iframe>';
             }
--- a/Classes/ViewHelpers/ExternalMediaViewHelper.php
+++ b/Classes/ViewHelpers/ExternalMediaViewHelper.php
@@ -29,6 +29,7 @@ class ExternalMediaViewHelper extends AbstractViewHelper
         $this->registerArgument('title', 'string', 'External media iframe title');
         $this->registerArgument('url', 'string', 'External media url');
         $this->registerArgument('class', 'string', 'CSS class rendered on the generated iframe code');
+        $this->registerArgument('referrerpolicy', 'string', 'Referrer policy rendered on the generated iframe code');
     }
 
     /**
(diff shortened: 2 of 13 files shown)


benjaminkott:
Thanks @Patta — the approach is right and the mechanism carries: `f:media`
folds `additionalAttributes` into the renderer options, and both
`YouTubeRenderer` and `VimeoRenderer` write them onto the iframe. Verified
against v13.4.34 and v14.3.6.

I rebased your commit onto `master` (the merge commit is gone — we rebase
rather than merge) and pushed three follow-ups on top. Yours is untouched.

**What the follow-ups fix**

`Media/Rendering/Video.html` does not only serve online media.
`Media/Type.html` routes by `bk2k:file.isMedia()`, which is true for every
extension in `SYS/mediafile_ext` — `mp4`, `webm`, `mov` among them — and a
`textmedia`, `tab` or `accordion` element hands those through unfiltered.
They resolve to `VideoTagRenderer`, which writes `additionalAttributes`
into the `<video>` tag. `referrerpolicy` is defined on `a`, `area`,
`iframe`, `img`, `link` and `script`, not on `video`, so every locally
hosted video came out with an attribute that is invalid there and does
nothing.

The partial now asks `OnlineMediaHelperRegistry` whether a helper is
registered for the file and hands the attribute over only then. An empty
setting renders no attribute at all instead of `referrerpolicy=""`.

Beyond that, the `external_media` element builds its iframe itself in
`ExternalMediaUtility` and had the same problem you described — it now
reads the same setting — and the setting carries a `description`, since
the manual generates its settings table straight out of
`settings.definitions.yaml` and there is no `enum` setting type on either
declared major.

`Tests/Functional/ContentElements/MediaTest.php` covers all three cases and
fails on two of them without the change to the partial.

PR #1638, opened 2026-08-24
[TASK] Declare package metadata in composer.json

benjaminkott:
Declares the package metadata TYPO3 v15 requires in `composer.json`, for this
extension and for the `demo_package` test fixture.

```json
"typo3/cms": {
    "extension-key": "bootstrap_package",
    "web-dir": ".build/public",
    "version": "16.0.0",
    "Package": {
        "providesPackages": {
            "scssphp/scssphp": "Contrib/scssphp"
        }
    }
}
```

### Why

From TYPO3 v15 the package manifest is read from `composer.json` alone.
`PackageManager::getComposerManifest()` throws `InvalidPackageManifestException`
(`1780502553`) for any extension that declares neither a version nor
`providesPackages` — classic mode installations, where `ext_emconf.php` used to
supply both. TYPO3 v14.2 already raises a deprecation for the missing fields
([Deprecation #108345](https://docs.typo3.org/permalink/changelog:deprecation-108345-1774126701));
it fires today in the functional suite (`PackageManager.php:948`).

`providesPackages` names the Composer packages the extension ships itself so
classic mode does not look for them in a vendor directory. That is
`scssphp/scssphp`, bundled under `Contrib/scssphp` (1.13.0) — the one entry in
`require` that is not a TYPO3 extension.

`ext_emconf.php` stays: Tailor and the TER read it.

### The dependency list moves with it

Declaring the two fields also stops TYPO3 v14 from evaluating `ext_emconf.php`,
and that includes the **dependency list**. The two manifests do not agree:

| | declares |
|---|---|
| `ext_emconf.php` `constraints.depends` | `typo3`, `rte_ckeditor`, `seo` |
| `composer.json` `require` | nine `typo3/cms-*` packages |

The functional test instances therefore have to load `extensionmanager` and
`install` on top of the `seo` and `rte_ckeditor` they already loaded, otherwise
the testing framework aborts with
`Package "bootstrap_package" depends on package "extensionmanager" which does not exist.`
That is what the change to the 22 test classes is.

**Worth a separate look:** nothing in `Classes/`, `Configuration/` or
`ext_localconf.php` references `TYPO3\CMS\Extensionmanager`, so
`typo3/cms-extensionmanager` looks like a stale `require`. Removing it is a
decision about the package's public dependency declaration, so it is not part of
this PR.

### Compatibility

TYPO3 v13.4 reads neither field. `PackageManager::mapExtensionManagerConfigurationToComposerManifest()`
overwrites `version` from `ext_emconf.php` unconditionally there, so the manifest
change is a no-op on v13.4. The extra `coreExtensionsToLoad` entries are
harmless on both versions.

### Keeping the version in step

`composer set-version` did not write `extra.typo3/cms.version`, so the new field
would have drifted on the next release.
[bk2k/extension-helper#6](https://github.com/benjaminkott/extension-helper/pull/6)
fixes that and is released as 2.2.0; the constraint is raised here accordingly.
`Tests/Unit/PackageManifestTest.php` asserts the two versions match and that every
entry in `providesPackages` is required and its directory exists.

### Checks

`composer test:php:lint`, `composer cgl:ci`, `composer phpstan`,
`composer test:php:unit`, `composer test:php:functional` — all green.

Part of a set of v13-compatible preparations that can be backported to
`BP_16_0` one by one.

### DIFF
--- a/Tests/Functional/Parser/ScssParserTest.php
+++ b/Tests/Functional/Parser/ScssParserTest.php
@@ -31,6 +31,8 @@ class ScssParserTest extends FunctionalTestCase
     protected array $coreExtensionsToLoad = [
         'seo',
         'rte_ckeditor',
+        'extensionmanager',
+        'install',
     ];
 
     protected array $testExtensionsToLoad = [
--- a/Tests/Functional/TypoScript/DynamicContentTest.php
+++ b/Tests/Functional/TypoScript/DynamicContentTest.php
@@ -27,6 +27,8 @@ final class DynamicContentTest extends FunctionalTestCase
     protected array $coreExtensionsToLoad = [
         'seo',
         'rte_ckeditor',
+        'extensionmanager',
+        'install',
     ];
 
     protected array $testExtensionsToLoad = [
--- a/Tests/Functional/Updates/AccordionContentElementUpdateTest.php
+++ b/Tests/Functional/Updates/AccordionContentElementUpdateTest.php
@@ -23,6 +23,8 @@ final class AccordionContentElementUpdateTest extends FunctionalTestCase
     protected array $coreExtensionsToLoad = [
         'seo',
         'rte_ckeditor',
+        'extensionmanager',
+        'install',
     ];
 
     protected array $testExtensionsToLoad = [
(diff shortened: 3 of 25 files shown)


benjaminkott:
Correction: the red `Build PHP (^14.3, 8.5)` on the first run was not infrastructure — it was this change. Declaring `version` and `providesPackages` stops TYPO3 v14 from evaluating `ext_emconf.php`, and that includes the dependency list. `composer.json` requires nine TYPO3 extensions where `ext_emconf.php` declared three, so the functional instances now need `extensionmanager` and `install` on top of `seo` and `rte_ckeditor`. Reproduced locally, fixed in the 22 test classes, whole suite green. The PR description is updated accordingly — my earlier claim that `composer test:php:functional` was green on this branch was wrong; only the unit filter had been run.

